PR #4, opened 2018-10-22
Introduce RemapImports module

axic:


libchisel/Cargo.toml
```diff
@@ -7,4 +7,6 @@ repository = "https://github.com/wasmx/wasm-chisel"
 homepage = "https://github.com/wasmx/wasm-chisel"
 
 [dependencies]
+log = "0.3"
```
jakelang: Where is this used?

libchisel/src/remapimports.rs
```diff
@@ -0,0 +1,124 @@
+use std::str;
```
jakelang: Generally this is implicit, not required. Same goes for `Vec` and `String` and other basic stdlib types.

libchisel/Cargo.toml
axic: Was used up until 077b7c411b7b6fa60b80a7d7b00b64e8d2abb005

libchisel/src/remapimports.rs
```diff
@@ -0,0 +1,124 @@
...
+fn import_section_mut(module: &mut Module) -> Option<&mut ImportSection> {
+    for section in module.sections_mut() {
+        if let &mut Section::Import(ref mut import_section) = section { return Some(import_section); }
+    }
+    None
+}
+
+pub fn rename_imports(module: &mut Module, translations: Translations) {
+    if let Some(section) = import_section_mut(module) {
+        for entry in section.entries_mut().iter_mut() {
```
jakelang: I think this can be simplified a bit into something like
`*section = section.entries_mut().iter_mut().map(|x| translations.get(x.module(), x.field()).collect()`
If a more functional style is preferred.

PR #5, opened 2018-10-22
Implement trimexports module

jakelang:


libchisel/src/trimexports.rs
```diff
@@ -0,0 +1,201 @@
...
+            .export()
+            .field("memory")
+            .internal()
+            .memory(0)
+            .build()
+            .build();
+
+        let trimmer = TrimExports::with_preset("ewasm");
+        let did_change = trimmer.translate(&mut module).unwrap();
+        assert!(!did_change);
```
axic: I think `assert_eq!(did_change, false)` is more readable here. The exclamation marks are too close.

libchisel/src/trimexports.rs
```diff
@@ -0,0 +1,201 @@
...
+            .signature()
+            .build()
+            .body()
+            .build()
+            .build()
+            .build();
+
+        let trimmer = TrimExports::with_preset("ewasm");
+        let did_change = trimmer.translate(&mut module).unwrap();
+        assert!(!did_change);
```
axic: Same here.

libchisel/src/trimexports.rs
```diff
@@ -0,0 +1,201 @@
...
+            .export()
+            .field("foo")
+            .internal()
+            .memory(0)
+            .build()
+            .build();
+
+        let trimmer = TrimExports::with_preset("ewasm");
+        let did_change = trimmer.translate(&mut module).unwrap();
+        assert!(did_change);
```
axic: For consistency then would use `assert_eq` here too.

PR #9, opened 2018-10-27
Validator module checking that a given function is exported

jakelang:


axic:
We have agreed on the last meeting that instead of this and #10 it would make sense to create a single general module taking a list of exports:
`checkexports(<list of exports>, allowUnlisted: bool)`

If `allowUnlisted` is true then it checks that whatever is requested is exports, but doesn't care if anything else is.

If `allowUnlisted` is false, the module can only export whatever is in the list.

And the list should have export name + export type/signature.

jakelang:
rebased.

jakelang:
> We have agreed on the last meeting that instead of this and #10 it would make sense to create a single general module taking a list of exports:
> `checkexports(<list of exports>, allowUnlisted: bool)`
> 
> If `allowUnlisted` is true then it checks that whatever is requested is exports, but doesn't care if anything else is.
> 
> If `allowUnlisted` is false, the module can only export whatever is in the list.
> 
> And the list should have export name + export type/signature.

according to my proposal this will be a macro-module calling this given module, with the said options

jakelang:
@axic I will do the macro-module in a separate PR. Best to merge this

axic:
> according to my proposal this will be a macro-module calling this given module, with the said options

I do not see a proposal here. Can you write it down into this issue?

jakelang:
```rust
//existing api
trait ModuleValidator {
	fn validate(self, module: &Module) -> Result<bool, String>;
}

//existing module. Can make a constructor to have `func` be a given value
struct CheckFuncExport {
	func: FunctionType, //signature to check for
}

//option 1: macro-module. maybe performance sacrifice, but less code reuse and smaller binaries
struct CheckMultipleFuncExports {
	funcs: Vec<FunctionType>,
}
impl ModuleValidator for CheckMultipleFuncExports {
		fn validate(self, module: &Module) -> Result<bool, String> {
				let mut ret = false;
				for function in self.funcs.iter() {
						let m = CheckFuncExport::new(func);
						ret |= m.validate(module);
				}
		}
}
```
copied from the hackmd

jakelang:
Closed in favor of #16

PR #12, opened 2018-10-28
Add very basic CLI

axic:


chisel/src/main.rs
```diff
@@ -0,0 +1,35 @@
...
+pub fn main() {
+    let args = env::args().collect::<Vec<_>>();
+    if args.len() != 3 {
+        println!("Usage: {} in.wasm out.wasm", args[0]);
+        return;
+    }
+
+    let mut file = File::open(&args[1]).expect("Failed to open file");
+    let mut code = Vec::new();
+    file.read_to_end(&mut code).expect("Failed to read file");
```
jakelang: You can simplify this using `std::fs::read(path: Path) -> Result<Vec<u8>>`
It also is faster because it pre-allocates the buffer before reading the file.

PR #11, opened 2018-10-28
Validator module checking that a memory is exported

jakelang:

PR #16, opened 2018-11-12
Implement VerifyExports

jakelang:
CheckExports essentially combines the functionality of the other export modules into one module that checks a list of exports, and for any extra exports.

Obsoletes #9 and #11.


libchisel/src/checkexports.rs
```diff
@@ -0,0 +1,177 @@
...
+            }
+        } else {
+            false
+        }
+    }
+}
+
+//NOTE: has_*_export is implemented with repeating code because you can't implement a trait for
+//enum variants, as they are not types. Furthermore, having one helper for non-func exports would
+//be ugly because information about the export type must still be passed down.
```
axic: Nitpicking: here comments don't have a space after the slash. Other comments use three slashes.

jakelang: 3 slash comments are for documentation using `rustdoc`. These are just in-source notes. I can make them all use a space, though.

libchisel/src/checkexports.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+}
+
+impl<'a> ModuleValidator for CheckExports<'a> {
+    fn validate(self, module: &Module) -> Result<bool, String> {
+        let required_exports_ok = !self
+            .entries
+            .iter()
+            .map(|x| x.is_exported(module))
```
jakelang: pls ignore shitty O(n^2) algo here.

libchisel/src/checkexports.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        let module_export_count = if let Some(section) = module.export_section() {
+            section.entries().len()
+        } else {
+            0
+        };
+
+        let extra_exports_ok = self.allow_unlisted || (self.entries.len() == module_export_count);
+
+        Ok(!required_exports_not_found && extra_exports_ok)
```
axic: I think these could be simplified:
```
let required_exports_ok = !required_exports_not_found;

if (!required_exports_ok)
  return Ok(false);

if (self.entries.len() != module_export_count)
  Ok(self.allow_unlisted)
```

jakelang: Yeah I prefer the localized return but this also works

axic: Yeah but it is hard to follow what's going on. I'm not sure the `extra_exports_ok` line is correct.

libchisel/src/checkexports.rs
```diff
@@ -0,0 +1,188 @@
...
+    } else {
+        false
+    }
+}
+
+/// Checks if a table is exported with the given name.
+fn has_table_export(section: &ExportSection, field: &str) -> bool {
+    if let Some(ref export) = section.entries().iter().find(|e| e.field() == field) {
+        match export.internal() {
+            Internal::Table(_index) => true,
```
axic: This is `_index` while the others above use `index`

PR #19, opened 2018-11-17
Introduce deployer module

axic:


libchisel/src/deployer.rs
```diff
@@ -0,0 +1,190 @@
...
+
+use super::*;
+
+use byteorder::{LittleEndian, WriteBytesExt};
+use rustc_hex::FromHex;
+
+fn deployer_code() -> Vec<u8> {
+    FromHex::from_hex(
+        "
+        0061736d010000000113046000017f60037f7f7f0060027f7f00600000023e0308
```
jakelang: Could you annotate with corresponding wast for clarity?

libchisel/src/deployer.rs
```diff
@@ -0,0 +1,190 @@
...
+
+    // Prepare payload (append length).
+    let mut custom_payload = payload.to_vec();
+    custom_payload
+        .write_i32::<LittleEndian>(payload.len() as i32)
+        .unwrap();
+
+    // Prepare and append custom section.
+    let custom = CustomSection {
+        name: "deployer".to_string(),
```
jakelang: Instead of having to fork an old version of parity, can we stick to the 0.21 tag and just implement a trait over `CustomSection` with a proper constructor? It would be much easier to maintain.

axic: I'd rather do it outside of this PR.

jakelang: If you implement a trait for a type, the trait has access to its type's members via `&self`

axic: Ah, so a trait _after_ the instance has been created.

jakelang: Not quite sure what you meant by that.

axic: Discussed this offline.

PR #20, opened 2018-11-18
Add ModuleCreator trait on deployer

axic:
Fixes #19.

### DIFF
--- a/libchisel/src/deployer.rs
+++ b/libchisel/src/deployer.rs
@@ -5,6 +5,33 @@ use super::*;
 use byteorder::{LittleEndian, WriteBytesExt};
 use rustc_hex::FromHex;
 
+pub enum DeployerType {
+    Memory,
+    CustomSection,
+}
+
+/// Struct on which ModuleCreator is implemented.
+pub struct Deployer {
+    pub deployer_type: DeployerType,
+    pub payload: Vec<u8>,
+}
+
+impl Deployer {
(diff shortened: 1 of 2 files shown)


axic:
Also need to add README changes.

libchisel/src/deployer.rs
```diff
@@ -5,6 +5,33 @@ use super::*;
 use byteorder::{LittleEndian, WriteBytesExt};
 use rustc_hex::FromHex;
 
+pub enum DeployerType {
+    Memory,
+    CustomSection,
+}
+
+/// Struct on which ModuleValidator is implemented.
```
jakelang: Documentation needs to be changed here for `ModuleCreator`

libchisel/src/deployer.rs
```diff
@@ -5,6 +5,33 @@ use super::*;
...
+pub struct Deployer {
+    pub deployer_type: DeployerType,
+    pub payload: Vec<u8>,
```
jakelang: I think it looks cleaner to use just the structured enum pattern, such that DeployerType variants will hold the payload.

axic: > I think it looks cleaner to use just the structured enum pattern, such that DeployerType variants will hold the payload.

Can you give an example how you imagine the API?

> Also, could you use a slice member with lifetime parameters instead of Vec?

Can you also give an example?

PR #29, opened 2018-11-21
Make VerifyExports Public

jakelang:


.gitignore
```diff
@@ -2,3 +2,4 @@
 **/*.rs.bk
 Cargo.lock
 **/*.swp
+**/chisel.yml
```
axic: Why was this needed? :)

jakelang:
that was a minor thing that was going to be merged in with cli pr anyway.
git status was being annoying

axic:
> that was a minor thing that was going to be merged in with cli pr anyway.
git status was being annoying

Why chisel cannot have a chisel configuration?

jakelang:
If we did want that I would make an examples dir.

axic:
> If we did want that I would make an examples dir.

Well even that is rejected by this gitignore entry.

But I mean an actual chisel file if chisel has a wasm target.

jakelang:
Oh yeah, in that case i will get rid of it.

jakelang:
> Oh yeah, in that case i will get rid of it.
In another PR when we make an examples dir.

PR #33, opened 2018-11-27
Add wasm_gc to chisel

### LINKED ISSUE #24: Add wrapper for wasm-gc
While wasm-gc was discontinued, due to Rust having LTO turned on, it is still useful for other languages besides Rust.

https://github.com/alexcrichton/wasm-gc

gballet:
Closes #24 
This is a first version of that PR, to gather initial comments. Still want to:

- [x] Use macros to reduce the amount of repeated code
- [ ] Write a unit test for each section
- [x] Save the `wasm_gc::Config` in the `WasmGC` structure

### DIFF
--- a/libchisel/src/gc.rs
+++ b/libchisel/src/gc.rs
@@ -1,0 +1,228 @@
+extern crate parity_wasm;
+extern crate wasm_gc;
+
+use super::ModuleTranslator;
+use parity_wasm::elements::*;
+
+pub struct WasmGC(wasm_gc::Config);
+
+impl Default for WasmGC {
+    fn default() -> WasmGC {
+        WasmGC(wasm_gc::Config::new())
+    }
+}
+
+impl WasmGC {
+    pub fn with_config(cfg: wasm_gc::Config) -> Self {
+        WasmGC(cfg)
+    }
+}
+
+macro_rules! update_section {
+    ($muts:expr, $cons:expr, $empty:expr) => {
+        match ($muts, $cons) {
+            (Some(m), Some(c)) => *m = c.clone(),
+            (Some(m), None) => *m = $empty,
+            _ => {}
+        };
+    };
+}
+
+impl ModuleTranslator for WasmGC {
+    fn translate(mut self, module: &mut Module) -> Result<bool, String> {
+        let serialized = parity_wasm::elements::serialize::<Module>(module.clone())
+            .expect("Could not serialize module");
+        match self.0.gc(&serialized[..]) {
+            Ok(gced_bytes) => {
+                let gced = parity_wasm::elements::deserialize_buffer::<Module>(&gced_bytes[..])
+                    .expect("Could not deserialize gc'ed module");
+
+                // Presumably, the custom section will not be modified
+
+                update_section!(
+                    module.type_section_mut(),
+                    gced.type_section(),
+                    TypeSection::with_types(vec![])
+                );
+                update_section!(
+                    module.import_section_mut(),
+                    gced.import_section(),
+                    ImportSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.function_section_mut(),
+                    gced.function_section(),
+                    FunctionSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.table_section_mut(),
+                    gced.table_section(),
+                    TableSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.memory_section_mut(),
+                    gced.memory_section(),
+                    MemorySection::with_entries(vec![])
+                );
+                update_section!(
+                    module.global_section_mut(),
+                    gced.global_section(),
+                    GlobalSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.export_section_mut(),
+                    gced.export_section(),
+                    ExportSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.elements_section_mut(),
+                    gced.elements_section(),
+                    ElementSection::with_entries(vec![])
+                );
+                update_section!(
+                    module.code_section_mut(),
+                    gced.code_section(),
+                    CodeSection::with_bodies(vec![])
+                );
+                update_section!(
+                    module.data_section_mut(),
+                    gced.data_section(),
+                    DataSection::with_entries(vec![])
+                );
+
+                if module.start_section() != gced.start_section() {
+                    if let Some(f) = gced.start_section() {
+                        module.set_start_section(f);
+                    } else {
+                        module.clear_start_section();
+                    }
+                }
+
+                Ok(gced_bytes.len() != serialized.len())
+            }
+            Err(e) => Err(format!("GC failure: {}", e)),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use parity_wasm::deserialize_buffer;
+    use rustc_hex::FromHex;
+
+    #[test]
+    fn do_not_touch_simple_module() {
+        let wasm: Vec<u8> = FromHex::from_hex("0061736d01000000").unwrap();
+
+        let mut module = deserialize_buffer::<Module>(&wasm).unwrap();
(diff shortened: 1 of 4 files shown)


libchisel/src/gc.rs
```diff
@@ -0,0 +1,150 @@
...
+extern crate parity_wasm;
+
+use super::ModuleTranslator;
+use parity_wasm::elements::*;
+
+pub struct WasmGC;
+
+impl WasmGC {
+    pub fn new() -> Self {
+        WasmGC{}
```
axic: Can use the `default` trait instead of this.

libchisel/src/gc.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+impl ModuleTranslator for WasmGC {
+    fn translate(self, module: &mut Module) -> Result<bool, String> {
+        let serialized = parity_wasm::elements::serialize::<Module>(module.clone()).expect("Could not serialize module");
+        match wasm_gc::Config::new().gc(&serialized[..]) {
+            Ok(gced_bytes) => {
+                let gced = parity_wasm::elements::deserialize_buffer::<Module>(&gced_bytes[..]).expect("Could not deserialize gc'ed module");
```
axic: Why not just return the modified version?

gballet: because the `Translator` interface doesn't return a module, and I am passed a mutable reference. We could change the interface to `Result<Module>` if it makes sense.

axic: Perhaps due to limitations in pwasm it would make sense to have two `translate` functions, one which returns a new instance and the current one which takes a mutable module.

The CLI then could try calling both and see which one succeeded. Or we could add helpers which rewrites a `&mut Module` with a given `Module`. (This is what @gballet is doing internally.)

@jakelang what do you think?

jakelang: I agree with this, or just choosing the new module route because handling lifetimes through this API makes for some inflexibility.

jakelang:
@gballet any idea why CI is not doing its job?

gballet:
Maybe it doesn't like me ?

jakelang:
@gballet let me know if this is ready for review by setting the `ready` label and removing `wip`. Also looking into why CI isn't working

jakelang:
@gballet Oh, CI is not running because you are opening a PR from your fork. @axic can add you so you can migrate this PR to a branch in this repo.

libchisel/src/gc.rs
axic: I still don't get why aren't we using this proposed API as opposed to this manual labour of changing mutable data?

libchisel/src/gc.rs
```diff
@@ -0,0 +1,119 @@
...
+        let result = WasmGC::default().translate(&mut module).unwrap();
+        assert_eq!(false, result);
+    }
+
+    #[test]
+    fn gc_does_something() {
+        let wasm: Vec<u8> = vec![0, 97, 115, 109, 1, 0, 0, 0, 1, 18, 4, 96, 1, 127, 1, 127, 96, 0, 1, 127, 96, 0, 0, 96, 1, 127, 1, 127, 2, 77, 5, 3, 101, 110, 118, 10, 109, 101, 109, 111, 114, 121, 66, 97, 115, 101, 3,127, 0, 3, 101, 110, 118, 6, 109, 101, 109, 111, 114, 121, 2, 0, 128, 2, 3, 101, 110, 118, 5, 116, 97, 98, 108, 101, 1, 112, 0, 0, 3, 101, 110, 118, 9, 116, 97, 98, 108, 101, 66, 97, 115, 101, 3, 127, 0, 3, 101, 110, 118, 5, 95, 112, 117, 116, 115, 0, 3, 3, 4, 3, 1, 2, 2, 6, 16, 3, 127, 1, 65, 0, 11, 127, 1, 65, 0, 11, 127, 0, 65, 0, 11, 7, 51, 4, 18, 95, 95, 112, 111, 115, 116,95, 105, 110, 115, 116, 97, 110, 116, 105, 97, 116, 101, 0, 3, 5, 95, 109, 97, 105, 110, 0, 1, 11, 114, 117, 110, 80, 111, 115, 116, 83, 101, 116, 115, 0, 2, 4, 95, 115, 116, 114, 3, 4, 9, 1, 0, 10, 37, 3, 9, 0, 35, 0, 16, 0, 26, 65, 0, 11, 3, 0, 1, 11, 21, 0, 35, 0, 65, 16, 106, 36, 2, 35, 2, 65, 128, 128, 192, 2, 106, 36, 3, 16, 2, 11, 11, 19, 1, 0, 35, 0, 11, 13, 104, 101, 108, 108, 111, 44, 32, 119, 111, 114, 108, 100, 33];
```
axic: Can all these just use `from_hex`? Easier to copy paste from files.

libchisel/src/gc.rs
```diff
@@ -0,0 +1,119 @@
...
+
+                update_section!(module.type_section_mut(), gced.type_section(), TypeSection::with_types(vec![]));
+                update_section!(module.import_section_mut(), gced.import_section(), ImportSection::with_entries(vec![]));
+                update_section!(module.function_section_mut(), gced.function_section(), FunctionSection::with_entries(vec![]));
+                update_section!(module.table_section_mut(), gced.table_section(), TableSection::with_entries(vec![]));
+                update_section!(module.memory_section_mut(), gced.memory_section(), MemorySection::with_entries(vec![]));
+                update_section!(module.global_section_mut(), gced.global_section(), GlobalSection::with_entries(vec![]));
+                update_section!(module.export_section_mut(), gced.export_section(), ExportSection::with_entries(vec![]));
+
+                // Awaiting mainlining of https://github.com/paritytech/parity-wasm/pull/250
```
axic: https://github.com/paritytech/parity-wasm/pull/250 was merged. Released in 0.35.3

libchisel/src/gc.rs
jakelang: decided offline: making this change after this is merged

libchisel/src/gc.rs
gballet: what makes it easier to read from file? I have my scripts that create those lists automatically.

axic: With `from_hex` there's no need for a script. However this is not crticial for this PR.

libchisel/src/gc.rs
axic: Please update the dependency to 0.35.3 and enable this feature

axic:
Please rebase and squash too.

libchisel/src/gc.rs
axic: In any case this implementation should be removed.

gballet:
Not done with all the feedback, I'll push the current state to check.

libchisel/src/gc.rs
gballet: yeah I think that returning a module is the best thing, because otherwise that's a lot of data copied in vain.

libchisel/src/gc.rs
gballet: `wasm_gc` pulls an older version and that is the one that is used, updating `Cargo.toml` has no effect.

libchisel/src/gc.rs
```diff
@@ -0,0 +1,228 @@
...
+extern crate wasm_gc;
+
+use super::ModuleTranslator;
+use parity_wasm::elements::*;
+
+pub struct WasmGC(wasm_gc::Config);
+
+impl Default for WasmGC {
+    fn default() -> WasmGC {
+        WasmGC(wasm_gc::Config::new())
```
axic: Why cannot we use the default trait here?

jakelang: Does `Config` implement or derive `Default`? I very rarely see `Default` used in the wild. Implementing `new` is much cleaner most of the time.

axic: Checked after I commented and it didn't.

gballet:
Closing this PR, it doesn't seem that it will be integrated.

Issue #18, opened 2018-11-17
WASM validator module


axic:

zixuanzh:
@axic do you mind linking to some references for this issue? thanks

axic:
This task has three parts:
1) Reviewing/documenting the semantic validation steps needed https://webassembly.github.io/spec/core/valid/index.html
2) Reviewing pwasm how much of that validation is implemented: https://github.com/paritytech/parity-wasm
3) Implementing the missing steps here.

zixuanzh:
@Steampunkery and I implemented verification for a subset of numeric instructions in our [fork]( https://github.com/zixuanzh/wasm-chisel/tree/verify-instructions). parity-wasm checks types for [const operators](https://github.com/paritytech/parity-wasm/blob/ce343b47d4dc7f00df7f4fbc6c54f8a6bcc168c1/src/elements/ops.rs#L163-L166) but not for binary as specified in the [spec](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-binop). 

We simulated the stack to check for types and hence we added a new validator interface that takes in a mutable self reference. We also use a self defined enum instead of String for error handling and propagation as advised by the [documentation](https://doc.rust-lang.org/book/first-edition/error-handling.html#defining-your-own-error-type). We will be happy to submit a PR if it will be helpful to wasm-chisel.

axic:
Can you please open a PR? We can review it that way.

pepyakin:
There is no validation in parity-wasm left and everything migrated to wasmi (see [this](https://github.com/paritytech/wasmi/blob/master/src/validation/func.rs)). Moreover, when we started compiling wasm into internal wasmi IR validation became mixed with compilation concerns.

[There was](https://github.com/paritytech/wasmi/issues/111) some interest in creating a separate reusable crate from third-parties. I think parity will need this as well at some point.

axic:
@pepyakin thanks for the heads up!

PR #37, opened 2018-12-06
Verify numeric instructions

zixuanzh:
As per issue #18 

parity-wasm checks types for [const operators](https://github.com/paritytech/parity-wasm/blob/ce343b47d4dc7f00df7f4fbc6c54f8a6bcc168c1/src/elements/ops.rs#L163-L166) but not for binary as specified in the [spec](https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-binop). 

We simulated the stack to check for types and hence we added a new validator interface that takes in a mutable self reference. We also use a self defined enum instead of String for error handling and propagation as advised by the [documentation](https://doc.rust-lang.org/book/first-edition/error-handling.html#defining-your-own-error-type).


libchisel/src/verifyinstructions.rs
```diff
@@ -0,0 +1,311 @@
...
+];
+
+
+use self::Filter::*;
+use std::mem::discriminant;
+
+type Pop = Vec<ValueType>;
+type Push = Vec<ValueType>;
+
+struct Signature {
```
jakelang: Can you document these structs with triple-slash comments? That way `rustdoc` can automatically generate a docs site for this.

libchisel/src/verifyinstructions.rs
```diff
@@ -0,0 +1,311 @@
...
+	pop: Pop,
+	push: Push
+}
+
+pub enum Filter {
```
jakelang: Perhaps it would be useful to have per-instruction granularity here?

libchisel/src/verifyinstructions.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    fn validate_instruction(
+        &mut self,
+        signature: &Signature,
+        instruction: &Instruction,
+    ) -> Result<bool, InstructionError> {
+        for signature_value in &signature.pop {
+            let value = self.stack.pop();
+            match value {
+                Some(stack_value) => {
```
jakelang: It is a matter of preference but I find it cleaner, in the case of destructuring two-valued enums like `Option`, to use the `if let Some(stack_value)` pattern here.

jakelang:
@zixuanzh CI seems to be a version behind on `rustfmt`. will fix that.

zixuanzh:
I am running on 1.32.0-nightly, fmt fails at deployer.rs etc in CI, it passes CI when I only commit verify-instructions related files.

jakelang:
@zixuanzh is this ready or WIP? Please add the respective label.

Also please rebase.

zixuanzh:
@jakelang it should be ready as it is unless @Steampunkery has something to add, we don't have access to add labels.

libchisel/src/lib.rs
```diff
@@ -35,6 +40,10 @@ pub trait ModuleValidator {
     fn validate(&self, module: &Module) -> Result<bool, ModuleError>;
 }
 
+pub trait InstructionValidator {
+    fn validate(&mut self, module: &Module) -> Result<bool, InstructionError>;
```
axic: I think since we have `ModuleError`, it should be possible to merge the `InstructionError` enum into it and thus removing the need for a new trait.

axic:
I think this looks good in general, great work @Steampunkery and @zixuanzh!

libchisel/src/instructionerrors.rs
```diff
@@ -0,0 +1,39 @@
...
+
+#[derive(Debug)]
+pub enum InstructionError {
+    GlobalNotFound,
+    LocalNotFound,
+    UnmatchedInstruction,
+    InvalidOperation(Instruction),
+}
+
+impl fmt::Display for InstructionError {
```
axic: If we merge `ModuleError` and `InstructionError`, could you create a new PR which just adds the `Display` and `Error` traits on `ModuleError`? Then this PR can just extend that with the new enum values.

jakelang: Alternatively, instead of adding new variants, it may make sense to just encode these errors as a string in the `Custom` variant of `ModuleError`.

jakelang: Or make `Custom(T)` type generic and throw `InstructionError` in there.

axic: Text in general is fine, but for programmatic detection the enums are better.

libchisel/src/verifyinstructions.rs
Steampunkery: Just pushed to the staging repo a moment ago, should show up in here soon. @zixuanzh  @jakelang

libchisel/src/instructionerrors.rs
Steampunkery: I'd like to stick to enum variants if possible, and I can certainly look into merging `InstructionError` into `ModuleError`.

axic:
@Steampunkery @zixuanzh can we take over this PR or do you think you'll have some resources to work on this in the next while?

zixuanzh:
@axic please proceed, we are unlikely to work on this in the immediate future. thanks

PR #41, opened 2018-12-06
Implement a non-destructive translate method

### LINKED ISSUE #34: Change ModuleTranslator to return a module rather than operate on a mutable reference
From https://github.com/wasmx/wasm-chisel/pull/33#discussion_r236871271

jakelang:
This second one returns a module as a result. The original method has been renamed to `translate_inplace`.
Closes #34

### DIFF
--- a/libchisel/src/lib.rs
+++ b/libchisel/src/lib.rs
@@ -16,7 +16,8 @@ pub trait ModuleCreator {
 }
 
 pub trait ModuleTranslator {
-    fn translate(self, module: &mut Module) -> Result<bool, String>;
+    fn translate(self, module: &Module) -> Result<Module, String>;
+    fn translate_inplace(self, module: &mut Module) -> Result<bool, String>;
 }
 
 pub trait ModuleValidator {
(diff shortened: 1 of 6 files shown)


libchisel/src/lib.rs
```diff
@@ -16,7 +16,8 @@ pub trait ModuleCreator {
 }
 
 pub trait ModuleTranslator {
-    fn translate(self, module: &mut Module) -> Result<bool, String>;
+    fn translate(self, module: &Module) -> Result<Module, String>;
+    fn translate_inplace(self, module: &mut Module) -> Result<bool, String>;
```
axic: Why is there a need for a different name?

jakelang: The alternative if you want to overload the method is put it in a different trait.

axic:
In the [original commet](https://github.com/wasmx/wasm-chisel/pull/33#discussion_r236871271) I said:
> The CLI then could try calling both and see which one succeeded.

Right now both methods do the exact same thing, because all the modules already implemented are capable to handle both cases. How do you plan to programmatically handle the case when the inplace version fails because it is unimplemented? `false` signals there was no changed made.

jakelang:
@axic The original comment was not clear at all about that. If a given module cannot implement one or the other, it should return `Err` with a string describing that case, for the caller to handle.
I am still not 100% sure what you mean by "not implemented". I am assuming you mean it cannot perform an in-place translation because of some issue with lifetimes, etc. In which case it should just have a dummy impl that returns `Err` as I said earlier.

axic:
It was likely discussed on Telegram.  We should aim to have the non-implemented case programmatically detectable. Expecting a random Error may or may not be the best way.

jakelang:
Perhaps we could just return a `ChiselError` enum wrapped in the usual `Result::Err`, which would make handling the error cases much cleaner. @axic

PR #52, opened 2018-12-17
Introduce ModuleError enum for errors 

axic:
Depends on #50.

### DIFF
--- a/libchisel/src/checkstartfunc.rs
+++ b/libchisel/src/checkstartfunc.rs
@@ -1,4 +1,4 @@
-use super::ModuleValidator;
+use super::{ModuleError, ModuleValidator};
 use parity_wasm::elements::Module;
 
 /// Struct on which ModuleValidator is implemented.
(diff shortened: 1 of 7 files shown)


libchisel/src/checkstartfunc.rs
```diff
@@ -15,7 +15,7 @@ impl CheckStartFunc {
 }
 
 impl ModuleValidator for CheckStartFunc {
-    fn validate(&self, module: &Module) -> Result<bool, String> {
+    fn validate(&self, module: &Module) -> Result<bool, ModuleError> {
```
jakelang: Alternative idea: Replace the current `Result<bool, ModuleError>` with a simpler `ModuleResult` which can have the variants `Valid, Invalid, Mutated, Unsupported, Other(String)`

jakelang: @axic what do you think about this?

axic: One downside is that I think `unwrap` cannot be used in that case.

jakelang: Oh yeah, true. I just realized that the methods are implemented directly rather than via a trait. this is good for merge then?

PR #56, opened 2018-12-21
Group with_preset methods into ModulePreset trait

jakelang:


libchisel/src/lib.rs
```diff
@@ -37,6 +37,12 @@ pub trait ModuleValidator {
     fn validate(&self, module: &Module) -> Result<bool, ModuleError>;
 }
 
+pub trait ModulePreset {
+    fn with_preset(&str) -> Result<Self, ()>
```
axic: I think we could have an `enum` for settings, which is then extended by the modules, just like how `ModuleError` works.

libchisel/src/trimexports.rs
```diff
@@ -40,6 +33,15 @@ impl ExportWhitelist {
             _ => Err(()),
         }
     }
+}
+
+impl ExportWhitelist {
```
axic: Why move this?

libchisel/src/verifyimports.rs
```diff
@@ -281,7 +281,9 @@ impl<'a> VerifyImports<'a> {
             _ => Err(()),
         }
     }
+}
 
+impl<'a> VerifyImports<'a> {
```
axic: Could wrap this entire implementation into `[cfg(test)]`, though if you do that, please do in a separate commit.

libchisel/src/trimexports.rs
jakelang: Out of habit I put trait impls before method impls.

libchisel/src/lib.rs
jakelang: Can you give an example of such a setting enum? I think that instantiating modules with the preset as a string is more robust when considering the possibility of third party modules as well as nonstandard presets.

libchisel/src/verifyimports.rs
jakelang: will do

libchisel/src/lib.rs
axic: `enum` for the optional settings, such as `payload` in `deployer`.

libchisel/src/verifyimports.rs
jakelang: Done

libchisel/src/lib.rs
jakelang: hmm. I'm starting to think that once configuration becomes a bit complex we should have a dedicated context structure

PR #59, opened 2018-12-31
Split the default ewasm import list out into its own module

jakelang:
Depends on #56 
Step 1 of `remapimports` refactor.

### DIFF
--- a/libchisel/src/imports.rs
+++ b/libchisel/src/imports.rs
@@ -1,0 +1,253 @@
+use super::ModulePreset;
+
+use parity_wasm::elements::{FunctionType, ValueType};
+
+pub struct ImportList<'a>(Vec<ImportType<'a>>);
+
+/// Enum internally representing a type of import.
+#[derive(Clone)]
+pub enum ImportType<'a> {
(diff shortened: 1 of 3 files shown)


axic:
Please don't merge this into `preset-trait`, changed the base to avoid that accident.

libchisel/src/imports.rs
```diff
@@ -0,0 +1,256 @@
...
+                ),
+                ImportType::Function(
+                    "ethereum",
+                    "selfDestruct",
+                    FunctionType::new(vec![ValueType::I32], None),
+                ),
+            ])),
+            //.iter()
+            //.cloned()
+            //.collect()
```
axic: Stray?

jakelang: Fixed.

jakelang:
@axic Is this good for merge? I have personally triple-checked the import list but we will introduce tests later.

axic:
It was approved, just I didn't take responsibility for double checking the list 😉

jakelang:
ok, merging

PR #62, opened 2019-01-02
[WIP] Add pwasm imports to verifyimports

axic:


axic:
I thought to push this half finish PR I had a while ago before somebody else duplicates work. Will rebase and change it to work with #59.

axic:
pwasm is deprecated, no need for this anymore.

PR #63, opened 2019-01-03
Change ModuleTranslator.translate() to return an Option<Module>

axic:
If it returns None that means no changes were made to the input.

### DIFF
--- a/libchisel/src/lib.rs
+++ b/libchisel/src/lib.rs
@@ -27,8 +27,8 @@ pub trait ModuleCreator {
 }
 
 pub trait ModuleTranslator {
-    /// Translates module. Returns new module. Can fail with ModuleError::NotSupported.
-    fn translate(&self, module: &Module) -> Result<Module, ModuleError>;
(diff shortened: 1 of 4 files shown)


libchisel/src/lib.rs
```diff
@@ -27,8 +27,8 @@ pub trait ModuleCreator {
 }
 
 pub trait ModuleTranslator {
-    /// Translates module. Returns new module. Can fail with ModuleError::NotSupported.
-    fn translate(&self, module: &Module) -> Result<Module, ModuleError>;
+    /// Translates module. Returns new module or none if nothing was modified. Can fail with ModuleError::NotSupported.
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError>;
```
jakelang: In the future I would like to find a better alternative to having the option nested in the result. De-structuring multiple layers becomes tedious.

axic: There's plenty of example of this. Not sure what is better, but it may premature optimisation.

PR #64, opened 2019-01-04
Use string with a lifetime in ModuleError::Custom

axic:


axic:
@jakelang I'm fine not merging it.

PR #66, opened 2019-01-09
Refactor RemapImports to use the common import list

### LINKED ISSUE #21: Cleanup remapimports to have a nicer structure

### LINKED ISSUE #31: Merge the EEI import list between verifyimports and renameimports

jakelang:
Closes #21.
Closes #31.


libchisel/src/verifyimports.rs
```diff
@@ -276,6 +276,7 @@ pub fn imported_func_sig_by_index(module: &Module, index: usize) -> FunctionType
     let import_section = module.import_section().expect("No function section found");
     let type_section = module.type_section().expect("No type section found");
 
+    assert!(index < import_section.entries().len());
```
axic: Can you place this into a separate commit and perhaps leave a note (not sure a note is actually needed).

jakelang: getting rid of this, actually. this is just leftover from bug hunting.

libchisel/src/imports.rs
```diff
@@ -13,6 +13,35 @@ pub enum ImportType<'a> {
     Table(&'a str, &'a str),
 }
 
+impl<'a> ImportType<'a> {
```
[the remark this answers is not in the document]
jakelang: These methods are super trivial. I don't think it is important to test these as opposed to other things.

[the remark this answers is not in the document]
axic: Some of them, like `look_up_by_field` don't seem to be trivial. The later we start adding tests for things the more burden it will feel like.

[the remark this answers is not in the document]
jakelang: Ok, will work on some tests. The top-level tests have smoke level coverage for it though.

libchisel/src/imports.rs
```diff
@@ -21,6 +50,28 @@ impl<'a> ImportList<'a> {
     pub fn with_entries(entries: Vec<ImportType<'a>>) -> Self {
         ImportList(entries)
     }
+
+    pub fn lookup_by_field(&self, name: &str) -> Option<&ImportType> {
+        let entries = self.entries();
+
+        for import in entries {
+            match import {
```
axic: Cannot this use `import.field() == name` ?

libchisel/src/remapimports.rs
```diff
@@ -1,242 +1,92 @@
...
 use parity_wasm::elements::*;
 
-#[derive(Hash, Eq, PartialEq, Debug, Clone)]
-pub struct ImportPair {
-    pub module: String,
-    pub field: String,
+pub struct RemapImports<'a> {
+    /// The list of correct imports.
+    list: ImportList<'a>,
+    /// The prefix that compiler-generated import names contain. In the case of ewasm, this is
```
axic: I think this is a bad direction. This module supposed to be very generic accepting a mapping from one import to another. The correct way to solve here if we don't want to have a big manual list is to have a helper which generates this prefixed version out of the official "ewasm" import list.

libchisel/src/imports.rs
jakelang: will refactor, this was written before the other helpers

libchisel/src/remapimports.rs
jakelang: Agreed. Although while this works for ewasm, I think it is wise to update this for 0.3.0 instead.

jakelang: @axic #70

axic: > Although while this works for ewasm, I think it is wise to update this for 0.3.0 instead.

Not sure what you mean. Do mean to first remove the functionality we have and then add it back?

libchisel/src/remapimports.rs
```diff
@@ -271,4 +123,109 @@ mod tests {
         assert_eq!(output, expected);
         assert!(did_change);
     }
+
+    #[test]
+    fn remap_did_mutate() {
```
axic: Are all these tests passing with the old version of the module?

jakelang: these were not present in the old version of the module. Will have to check.

libchisel/src/remapimports.rs
jakelang: I mean that this works for the current use cases. We can make the prefix feature optional or remove it later.

libchisel/src/remapimports.rs
axic: Tested locally. They do work with the old code. Can we create a PR to merge tests in?

axic: Done in #80.

axic:
@jakelang please rebase

libchisel/src/remapimports.rs
axic: For eth2 this is now badly needed.

libchisel/src/imports.rs
```diff
@@ -142,6 +182,16 @@ impl<'a> ModulePreset for ImportList<'a> {
                     "getExternalCodeSize",
                     FunctionType::new(vec![ValueType::I32], Some(ValueType::I32)),
                 ),
+                ImportType::Function(
+                    "ethereum",
+                    "getExternalCodeSize",
```
axic: This is a duplicate of above.

libchisel/src/imports.rs
```diff
@@ -142,6 +182,16 @@ impl<'a> ModulePreset for ImportList<'a> {
...
+                    FunctionType::new(vec![ValueType::I32], Some(ValueType::I32)),
+                ),
+                ImportType::Function(
+                    "ethereum",
+                    "codeCopy",
```
axic: This is a duplicate below.

libchisel/src/imports.rs
```diff
@@ -13,6 +13,35 @@ pub enum ImportType<'a> {
     Table(&'a str, &'a str),
 }
 
+impl<'a> ImportType<'a> {
+    pub fn module(&self) -> &'a str {
+        // FIXME: Is there a way to shorten this expression?
```
axic: I guess upstreaming a change?

PR #67, opened 2019-01-10
Fix indexing in VerifyImports

### LINKED ISSUE #47: VerifyImports fails on code that has been transformed by RemapImports
For example:
```wat
(module
  (memory 1)
  (type (;0;) (func (result i64)))
  (import "ethereum" "getGasLeft" (type 0))
  (func $main)
  (export "main" (func $main))
  (export "memory" (memory 0))
)
```

jakelang:
Closes #47.

### DIFF
--- a/libchisel/src/verifyimports.rs
+++ b/libchisel/src/verifyimports.rs
@@ -160,14 +160,10 @@ impl<'a> ImportCheck for ImportType<'a> {
                 .find(|e| e.field() == *field_str && *module_str == e.module())
             {
                 match entry.external() {
-                    // TODO: Wrap this in a helper.
                     External::Function(idx) => {
-                        if let Some(sig) = func_sig {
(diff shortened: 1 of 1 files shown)


libchisel/src/verifyimports.rs
```diff
@@ -169,7 +169,7 @@ impl<'a> ImportCheck for ImportType<'a> {
                                 ImportStatus::Malformed
                             }
                         } else {
-                            ImportStatus::Malformed
+                            panic!(); // This should never happen.
```
axic: Instead of `panic!()` here cannot you just do `let sig = func_sig.unwrap()`?

axic: Or even `.expect`

jakelang: Will do

PR #71, opened 2019-01-14
Add From<String> trait on ModuleError

axic:


libchisel/src/lib.rs
```diff
@@ -128,6 +134,12 @@ mod tests {
         assert!(result.is_ok());
     }
 
+    #[test]
+    fn from_error() {
+        let err: ModuleError = "custom message".to_string().into();
```
axic: Not really useful in this form, but is when trying to unwrap from a function returning `Result<..., String>` into `Result<..., ModuleError>`

libchisel/src/lib.rs
```diff
@@ -45,6 +45,12 @@ pub trait ModulePreset {
         Self: std::marker::Sized;
 }
 
+impl From<String> for ModuleError {
```
jakelang: Could perhaps also implement `From<&'a str>` as well.

axic: Do you want to implement that in a new PR? I'm for just merging this.

jakelang: ok, sounds good

Issue #48, opened 2018-12-17
Translator to drop sections


axic:

axic:
Some analysis can be found here: https://github.com/ewasm/ewasm-precompiles/pull/32#issuecomment-449574040

axic:
Need to still expose this in the CLI.

PR #72, opened 2019-01-14
Introduce dropsection 

axic:
Implements #48.

### DIFF
--- a/libchisel/src/dropsection.rs
+++ b/libchisel/src/dropsection.rs
@@ -1,0 +1,116 @@
+use super::{ModuleError, ModuleTranslator};
+
+use parity_wasm::builder::*;
+use parity_wasm::elements::*;
+
+/// Enum on which ModuleTranslator is implemented.
+pub enum DropSection<'a> {
+    NamesSection,
+    /// Name of the custom section.
+    CustomSectionByName(&'a String),
+    /// Index of the custom section.
+    CustomSectionByIndex(usize),
+    /// Index of the unknown section.
+    UnknownSectionByIndex(usize),
+}
+
+fn names_section_index_for(module: &Module) -> Result<usize, String> {
+    for (index, section) in module.sections().iter().enumerate() {
+        if let Section::Name(_section) = section {
+            return Ok(index);
+        }
+    }
+    Err("Not found".to_string())
+}
+
+fn custom_section_index_for(module: &Module, name: &String) -> Result<usize, String> {
+    for (index, section) in module.sections().iter().enumerate() {
+        if let Section::Custom(_section) = section {
+            if _section.name() == name {
+                return Ok(index);
+            }
+        }
+    }
+    Err("Not found".to_string())
+}
+
+impl<'a> DropSection<'a> {
+    fn find_index(&self, module: &Module) -> Result<usize, String> {
+        Ok(match &self {
+            DropSection::NamesSection => names_section_index_for(module)?,
+            DropSection::CustomSectionByName(name) => custom_section_index_for(module, &name)?,
+            DropSection::CustomSectionByIndex(index) => *index,
+            DropSection::UnknownSectionByIndex(index) => *index,
+        })
+    }
+
+    fn drop_section(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        let index = self.find_index(&module)?;
+
+        let sections = module.sections_mut();
+        if index > sections.len() {
+            return Err(ModuleError::NotFound);
+        }
+        sections.remove(index);
+
+        Ok(true)
+    }
+}
+
+impl<'a> ModuleTranslator for DropSection<'a> {
+    fn translate_inplace(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        Ok(self.drop_section(module)?)
+    }
+
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError> {
+        let mut ret = module.clone();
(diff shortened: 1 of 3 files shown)


libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,116 @@
+use super::{ModuleError, ModuleTranslator};
+
+use parity_wasm::builder::*;
+use parity_wasm::elements::*;
+
+/// Enum on which ModuleTranslator is implemented.
+pub enum DropSection<'a> {
```
axic: Probably will rewrite to have this as `DropSectionKind` and `DropSection` as a vector of `DropSectionKind`. In that case the entire code can be rewritten somewhat, potentially with a single loop, though one must be careful when removing sections with a lower index.

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,116 @@
...
+    NamesSection,
+    /// Name of the custom section.
+    CustomSectionByName(&'a String),
+    /// Index of the custom section.
+    CustomSectionByIndex(usize),
+    /// Index of the unknown section.
+    UnknownSectionByIndex(usize),
+}
+
+fn names_section_index_for(module: &Module) -> Result<usize, String> {
```
axic: These can be changed into `Option`. Or could introduce `ModuleError::NotFound`. (Actually did that in a commit, but haven't pushed.)

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,116 @@
...
+            DropSection::UnknownSectionByIndex(index) => *index,
+        })
+    }
+
+    fn drop_section(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        let index = self.find_index(&module)?;
+
+        let sections = module.sections_mut();
+        if index > sections.len() {
+            return Err(ModuleError::Custom("Not found.".to_string()));
```
axic: Not sure if this should be considered an error or marked as `Ok(false)`.

Potentially we need a flag in the translator configuration saying whether fields not found is an error or not and behave accordingly.

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+    }
+    Err("Not found".to_string())
+}
+
+impl<'a> DropSection<'a> {
+    fn find_index(&self, module: &Module) -> Result<usize, String> {
+        Ok(match &self {
+            DropSection::NamesSection => names_section_index_for(module)?,
```
axic: This was named `_for` because it had a section enum as a parameter to be generic. If we keep these helpers, they need better naming.

libchisel/src/dropsection.rs
axic: After this rewrite I could also introduce `with_preset` with `ewasm` which removes the namessection and all custom sections with names `debug`.

libchisel/src/dropsection.rs
jakelang: Since there is no case where `Ok(false)` is returned and `false` indicates no mutation, it probably makes sense that this should be OK. Although a flag would make it more versatile.

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,116 @@
+use super::{ModuleError, ModuleTranslator};
+
+use parity_wasm::builder::*;
+use parity_wasm::elements::*;
+
+/// Enum on which ModuleTranslator is implemented.
+pub enum DropSection<'a> {
+    NamesSection,
+    /// Name of the custom section.
+    CustomSectionByName(&'a String),
```
jakelang: Perhaps another mode for DropSection would be identifying data sections that are used for rust debugging and panic messages. 
Identifying these is not exactly straightforward, but one option is just using a plain regex lookup for things like `libcore` in the data sections. Not sure what the risk of false positives for this would be.

axic:
@jakelang where do we stand with this?

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    #[test]
+    fn remove_oob_unknown_section_by_index() {
+        let mut module = builder::module()
+            .with_section(Section::Custom(CustomSection::new(
+                "test".to_string(),
+                vec![],
+            )))
+            .build();
+        let dropper = DropSection::UnknownSectionByIndex(1);
```
jakelang: Maybe annotate why this would not mutate.

libchisel/src/dropsection.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Name of the custom section.
+    CustomSectionByName(&'a String),
+    /// Index of the custom section.
+    CustomSectionByIndex(usize),
+    /// Index of the unknown section.
+    UnknownSectionByIndex(usize),
+}
+
+fn names_section_index_for(module: &Module) -> Option<usize> {
+    for (index, section) in module.sections().iter().enumerate() {
```
jakelang: You can iterate over it and use `position` to find the section in a more concise way. For example:
`module.sections().iter().position(|e| match e { Section::Name(_) => true, _ => false })`

libchisel/src/dropsection.rs
axic: The test name says "oob_unknown_section_index" where oob means out of bounds. Is that enough?

jakelang: It is but at a glance it would be nicer to have some comments or something in the code. It's fine

Issue #76, opened 2019-01-21
Introduce IR for analysis


axic:
Or at least helpers with some data structures.

A good start would be doing a CFG (control flow graph) or at least a primitive version which show dependencies of functions.

This can be used in two ways in the short term:
- finding unused functions
- in "atomic metering" to find the leaf nodes to start metering on

axic:
Could consider using https://github.com/rustwasm/walrus because it has the same goals.

PR #77, opened 2019-01-25
Introduce helpers for building function dependency graphs

jakelang:
Part of #76.


libchisel/src/funcdeps.rs
```diff
@@ -0,0 +1,83 @@
...
+/// A function dependency graph is represented as a list of "edges", or pairs of function indices
+/// (a, b) where a calls b.
+
+/// An edge, where the function at the left index calls the function at the right
+/// index.
+#[derive(PartialEq, Eq, Hash)]
+struct Edge(u32, u32);
+
+/// Container struct for the function dependency graph.
+pub struct DepGraph {
```
axic: Please be consistent in naming the file and this struct. Just choose one.

libchisel/src/funcdeps.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// Private interface for managing the function dependency graph
+trait GraphManager {
+    fn probe(&mut self, idx: u32, bodies: &[FuncBody]);
+    fn add_edge(&mut self, dep: Edge) -> bool;
+}
+
+/// Public interface for building function dependency graphs.
+pub trait GraphBuilder: GraphManager {
+    fn build(module: &Module, entry: u32) -> Result<Self, ()>
```
axic: Should document `entry` here is a function index.

axic: Should also create some helpers outside which translate a function name from `NamesSection` into an index.

libchisel/src/funcdeps.rs
```diff
@@ -0,0 +1,83 @@
...
+            edges: HashSet::new(),
+        }
+    }
+}
+
+impl GraphManager for DepGraph {
+    /// Recursively searches function bodies for calls to other functions and adds edges
+    /// accordingly.
+    fn probe(&mut self, idx: u32, bodies: &[FuncBody]) {
+        if (idx as usize) < bodies.len() {
```
axic: Is it `u32` in the wasm spec?

axic: Could also just use assert if we do panic already.

libchisel/src/funcdeps.rs
```diff
@@ -0,0 +1,83 @@
...
+            let func_body = &bodies[idx as usize];
```
axic: Isn't there a helper in pwasm for this?

libchisel/src/funcdeps.rs
jakelang: Fixed.

libchisel/src/funcdeps.rs
jakelang: > Should also create some helpers outside which translate a function name from `Exports`/`NamesSection` into an index.

This is the plan.



> Should document `entry` here is a function index.

fixed. For some reason the trait impl is allowed to have different arg names than the trait definition.

libchisel/src/funcdeps.rs
jakelang: > Is it `u32` in the wasm spec?

It is in the wasm spec. The `usize` casts are there because pwasm and because usize is the only valid type for the indexing operator.

libchisel/src/funcdeps.rs
jakelang: I did not see one. Please link if there is.

PR #85, opened 2019-04-17
Grow deployer's memory to size

### LINKED ISSUE #53: Deployer: fix memory size
The deployer has a fixed 1 page memory size (this applies to both the memory and the custom section deployre). Likely it will fail with larger contacts. This needs to be fixed and tested for.

s1na:
Fixes #53

### DIFF
--- a/libchisel/src/deployer.rs
+++ b/libchisel/src/deployer.rs
@@ -75,6 +75,11 @@ fn create_custom_deployer(payload: &[u8]) -> Module {
     // This is the pre-written deployer code.
     let mut module: Module = parity_wasm::deserialize_buffer(&code).expect("Failed to load module");
 
+    // Re-write memory to pre-allocate enough for code size
+    let memory_initial = (payload.len() as u32 / 65536) + 1;
+    let mem_type = parity_wasm::elements::MemoryType::new(memory_initial, None, false);
+    module.memory_section_mut().unwrap().entries_mut()[0] = mem_type;
+
     // Prepare payload (append length).
     let mut custom_payload = payload.to_vec();
     custom_payload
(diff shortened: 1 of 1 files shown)


libchisel/src/deployer.rs
```diff
@@ -38,6 +38,9 @@ impl<'a> Deployer<'a> {
     (local $payload_size i32)
     (set_local $size (call $getCodeSize))
 
+    ;; grow memory to fit code; assumes 1 initial page
+    (call $growToSize (get_local $size))
```
axic: I think instead of this we may be better off using the module builder and then we can just change `(memory 1)`. What do you think?

libchisel/src/deployer.rs
```diff
@@ -50,17 +53,30 @@ impl<'a> Deployer<'a> {
...
   )
+
+  (func $growToSize (param i32)
+    (local $pages i32)
+    ;; pages = size / 2^16
+    ;; assumes current memory size is 1 page
+    (set_local $pages (i32.shr_u (get_local 0) (i32.const 16)))
+    ;; throw if growing memory failed
+    (if
+      (i32.eq (memory.grow (get_local $pages)) (i32.const -1))
```
axic: Why not just `i32.const 65536` instead of this pages stuff?

libchisel/src/deployer.rs
s1na: That makes more sense, it'll also have less runtime cost. Will update.

s1na: Couldn't use builder to modify existing memory. Modified the module in-place.

axic:
Nice! 

Can you also update the memory deployer? You only did the custom section deployer.

s1na:
Sorry, updated. 

I kind of had the wrong impression that wasm engines would extend memory for the data section...

axic:
> I kind of had the wrong impression that wasm engines would extend memory for the data section...

Tbh I'm not sure, need to validate both cases. But having the bigger memory explicitly cannot hurt.

PR #86, opened 2019-04-30
add Dockerfile for chisel

jwasinger:
Doesn't use the `ewasm/rust-wasm` docker image because the extra tooling isn't needed.


jwasinger:
Yeah, let's do that.

PR #87, opened 2019-04-30
Module for removing start function (trimstartfunction)

hugo-dc:
Added code to trim the start function if present.

Here is an example of a chisel.yml using this functionality:

```yaml
ewasm:
  file: "build/optimized.wasm"
  output: "ewasm_token.wasm"
  trimstartfunc:
    preset: "ewasm"
```

The preset value is not being used, what would be the right configuration in  `chisel.yml`?
I was thinking something like this:

```yaml
ewasm:
  file: "build/optimized.wasm"
  output: "ewasm_token.wasm"
  trimstartfunc: true
```

@jakelang please take a look and let me know your comments.

### DIFF
--- a/chisel/src/main.rs
+++ b/chisel/src/main.rs
@@ -10,8 +10,8 @@ use std::fs::{read, read_to_string};
 use std::process;
 
 use libchisel::{
-    checkstartfunc::*, deployer::*, remapimports::*, trimexports::*, verifyexports::*,
-    verifyimports::*,
+    checkstartfunc::*, deployer::*, remapimports::*, trimexports::*, trimstartfunc::*,
+    verifyexports::*, verifyimports::*,
 };
 
 use clap::{App, Arg, ArgMatches, SubCommand};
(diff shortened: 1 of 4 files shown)


chisel/src/main.rs
```diff
@@ -224,6 +224,14 @@ fn execute_module(context: &ModuleContext, module: &mut Module) -> bool {
                 Err("trimexports: Invalid preset")
             }
         }
+        "trimstartfunc" => {
+            if let Some(start_section) = module.start_section() {
```
axic: Can you please create a module for this? See #88 for an example of a tiny module.

There are two reasons:
- chisel is only a frontend, perhaps others want to use libchisel directly
- we can add test cases in libchisel

hugo-dc: @axic thanks for the comments, moved to a new module and added some test cases.

hugo-dc: pending to add documentation, still not sure about how it should be configured in chisel.yml because this should work the same for "ewasm" or "pwasm".

chisel/src/main.rs
```diff
@@ -218,12 +218,18 @@ fn execute_module(context: &ModuleContext, module: &mut Module) -> bool {
...
+
             if let Ok(chisel) = TrimExports::with_preset(&preset) {
                 translate_module(module, chisel)
             } else {
                 Err("trimexports: Invalid preset")
             }
         }
+        "trimstartfunc" => {
+            is_translator = true;
+            let chisel = TrimStartFunc::new();
```
jakelang: You can make preset-style options for a chisel module by implementing `ModulePreset`.
I don't really think it is needed here, but it looks cleaner until I finally clean up the config loader.

hugo-dc: thanks @jakelang , implemented `ModulePreset`

hugo-dc:
> Can you put in the output expectation?

you mean, in the documentation/README.md? I added some comments to the README.md file, please let me know if that is what you are referring to.

still pending to squash commits.

axic:
> Can you put in the output expectation?

I meant to compare the resulting bytecode to the expected bytecode.

hugo-dc:
changed to compare final vs initial bytecode

axic:
Thanks!

PR #88, opened 2019-04-30
Introduce repack module

axic:
Inspired by https://github.com/ewasm/sentinel-rs/pull/7 from @cdetrio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,10 @@ Wraps module into an ewasm-compatible constructor. It has two presets:
 - `memory`: wrap the module as a pre-defined memory section
 - `customsection`: include the module as a custom section
 
+### repack
+
+Re-serializes the module. It will drop any unknown (custom) sections.
+
 ## CLI
 
 `chisel` is available as a command line tool.
--- a/libchisel/src/repack.rs
+++ b/libchisel/src/repack.rs
@@ -1,0 +1,93 @@
+use super::{ModuleError, ModuleTranslator};
+use parity_wasm::builder;
+use parity_wasm::elements::*;
+
+pub struct Repack;
+
+impl Repack {
+    pub fn new() -> Self {
+        Repack {}
+    }
+}
+
+impl ModuleTranslator for Repack {
+    fn translate_inplace(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError> {
+        // TODO: check in names section is carried over.
+        let module = module.clone();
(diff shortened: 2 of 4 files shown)


axic:
Need to finish the test cases.

libchisel/src/repack.rs
```diff
@@ -0,0 +1,124 @@
...
+            .build()
+            .build()
+            .export()
+            .field("_call")
+            .internal()
+            .func(0)
+            .build()
+            .build();
+
+        let trimmer = TrimExports::with_preset("pwasm").unwrap();
```
jakelang: what's going on here with these tests?
From what I gather this should be repack.

README.md
```diff
@@ -43,6 +43,10 @@ Wraps module into an ewasm-compatible constructor. It has two presets:
 - `memory`: wrap the module as a pre-defined memory section
 - `customsection`: include the module as a custom section
 
+### repack
+
+Re-serializes the module. It will drop any unknown (custom) sections.
```
jakelang: I think the wording in the first sentence is a bit unintuitive. Maybe the second sentence on its own will do.

libchisel/src/repack.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+impl ModuleTranslator for Repack {
+    fn translate_inplace(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError> {
+        let module = module.clone();
+        let module = builder::from_module(module).build();
```
jakelang: Are we certain that this drops custom sections?

libchisel/src/repack.rs
axic: As I said above tests need to be finished. The source file was trimexports.

libchisel/src/repack.rs
axic: Based on empirical tests it drops everything.

axic:
> Maybe would be better to include this as an option for DropSection?

Yeah when dropsection is finished, but this works already?

libchisel/src/repack.rs
jakelang: everything being all "unknown" sections?

axic: It dropped the names section too. Technically it just drops every unparsed section and none of the custom sections are parsed by default by pwasm.

README.md
axic: Well the second sentence is not only what it does, it is a side effect.

axic:
Merging this now as it is a need in for other projects (scout, ewasm-precompiles).

PR #89, opened 2019-04-30
validates the 'index' namespace (generated by assemblyscript)

hugo-dc:
When I was testing the WRC20 implementation in AssemblyScript I realized it is generated using the "index" namespace instead of "env", when I run chisel to the resulting wasm file, it says everything is fine and doesn't remap the imports (the index namespace remains).

This PR adds the "index" namespace to be replaced by the "ethereum" namespace.

### DIFF
--- a/libchisel/src/remapimports.rs
+++ b/libchisel/src/remapimports.rs
@@ -28,39 +28,41 @@ impl ModulePreset for Translations {
         match preset {
             "ewasm" => {
                 let namespaces = vec!["env", "index"];
-                let methods = vec!["useGas",
-                                   "getGasLeft",
-                                   "getAddress",
-                                   "getBlockHash",
-                                   "getBlockCoinbase",
-                                   "getBlockDifficulty",
-                                   "getBlockGasLimit",
-                                   "getBlockNumber",
-                                   "getBlockTimestamp",
-                                   "getExternalBalance",
-                                   "getTxGasPrice",
-                                   "getTxOrigin",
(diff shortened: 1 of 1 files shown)


libchisel/src/remapimports.rs
```diff
@@ -27,143 +27,56 @@ impl ModulePreset for Translations {
...
-                    ),
-                    (
-                        ImportPair::new("env", "ethereum_selfDestruct"),
-                        ImportPair::new("ethereum", "selfDestruct"),
-                    ),
-                ]
-                .iter()
-                .cloned()
-                .collect();
+                let namespaces = vec!["env", "index"];
```
axic: I think Assemblyscript has a decorator to change the namespace. 

See [this](https://github.com/ewasm/assemblyscript-ewasm-api/pull/2/files#diff-03ea643b15c0986ca110e293df93a9d6R76):
```
@external("debug", "print64")
export declare function print64(value: i64): void;
```

hugo-dc: @axic you're right, for AssemblyScript you can use the decorator `@external("ethereum", "methodName")`

axic: Can we close this PR then?

It would be nice starting a guide on languages, perhaps in the `ewasm/design` repo for now.

hugo-dc:
I'm ok closing this PR, and have a guide on languages in the `ewasm/design` repo instead.

PR #90, opened 2019-05-01
CLI: return a zero exit code when a module is not modified during translation

jwasinger:
If I was using chisel in some kind of build pipeline and module translation resulted in no changes to the wasm, that would currently be interpreted as an error.


jwasinger:
hey @jakelang what do you think about this?

jakelang:
Looks good. In general, I want to adhere more to posix exit codes like this.

PR #92, opened 2019-05-17
add 'ewasm_clang' preset to remapimports

jwasinger:
Allows chisel to support C contracts (other than the issue of the entrypoint being exported as `_main` which will be dealt with in it's own PR).

### DIFF
--- a/libchisel/src/remapimports.rs
+++ b/libchisel/src/remapimports.rs
@@ -167,7 +167,149 @@ impl ModulePreset for Translations {
                 Ok(Translations {
                     translations: trans,
                 })
-            }
+            },
+            "ewasm_clang" => {
+                let trans: HashMap<ImportPair, ImportPair> = [
+                    (
+                        ImportPair::new("env", "useGas"),
+                        ImportPair::new("ethereum", "useGas"),
+                    ),
+                    (
+                        ImportPair::new("env", "getGasLeft"),
+                        ImportPair::new("ethereum", "getGasLeft"),
+                    ),
+                    (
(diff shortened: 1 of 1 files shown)


jwasinger:
will add a small unit test to this.

axic:
To be honest I still don't get why we can't agree to use a single notation. `ewasm_clang` is a misleading name though.

We shouldn't merge this before #66 though.

jwasinger:
Any suggestions on a better name?  I'm not really sure myself.

> To be honest I still don't get why we can't agree to use a single notation

What do you mean?

axic:
> > To be honest I still don't get why we can't agree to use a single notation
>
> What do you mean?

To use a single naming structure. Most of the tools use the `ethereum_` prefix on the EEI to avoid naming clashes in the target language.

jwasinger:
It seems C doesn't support modules for function imports and defaults to "env" no matter what.  So this is a bit of a hack to get around C llvm wasm tooling being underdeveloped.

axic:
> It seems C doesn't support modules for function imports and defaults to "env" no matter what. So this is a bit of a hack to get around C llvm wasm tooling being underdeveloped.

The entire point of `remapimports` is that most languages don't support the namespace, but default to `env.` This is what `remapimports` is doing, but expects that the function names are prefixed with `ethereum_`.

E.g. `ethereum_getBalance` in Rust is translated into `env` `ethereum_getBalance` in wasm. We translate that to `ethereum` `getBalance` in chisel.

PR #94, opened 2019-05-27
THE BIG CLI REWRITE

jakelang:
WE ARE A MODERN INDEPENDENT SOFTWARE PROJECT SO LETS HAVE THE UI OF ONE


axic:
@jakelang is this going to be finished?

chisel/src/result.rs
```diff
@@ -0,0 +1,211 @@
...
+use parity_wasm::elements::{serialize, serialize_to_file, Module};
+
+#[derive(Clone)]
+/// Main result structure returned by ChiselDriver, containing a manifest of modules executed and
+/// exposing methods to write output from translators and creators.
+pub struct ChiselResult(Vec<RSResult>);
+
+#[derive(Clone)]
+/// Ruleset result structure.
+pub struct RSResult {
```
axic: Why not `RulesetResult`?

chisel/src/result.rs
```diff
@@ -0,0 +1,211 @@
...
+            }
+        } else {
+            Ok(false)
+        }
+    }
+
+    #[cfg(feature = "wabt-utils")]
+    pub fn write_output_to_file_wat(&self) -> Result<bool, Box<dyn Error>> {
+        if let Some(ref module) = self.output_module {
+            match serialize(module.clone()) {
```
axic: Use `module.clone().to_bytes()` here.

chisel/src/result.rs
```diff
@@ -0,0 +1,221 @@
...
+                Err(e) => Err(e.into()),
+            }
+        } else {
+            Ok(false)
+        }
+    }
+
+    pub fn write_output_to_file_hex(&self) -> Result<bool, Box<dyn Error>> {
+        if let Some(ref module) = self.output_module {
+            // TODO: Get rid of module clone here
```
axic: If you want to get rid of the clone change `write_output_file_hex` to not take a ref.

chisel/src/main.rs
```diff
@@ -8,400 +8,110 @@ extern crate serde_yaml;
...
-                .long("verbose")
-                .help("Enables verbose debug logging")
+            Arg::with_name("NO_RECOVER")
+                .short("x")
+                .long("norecover")
+                .help("Exits immediately on all recoverable errors")
                 .global(true),
         )
+        .arg(
+            Arg::with_name("DEBUG_MESSAGES")
```
axic: Why not keep `VERBOSE` as before?

chisel/src/options.rs
```diff
@@ -0,0 +1,118 @@
...
+
+impl Default for ChiselFlags {
+    fn default() -> Self {
+        let opts: HashMap<String, String> = HashMap::new();
+        let mut ret = ChiselFlags(opts);
+
+        ret.set("util.norecover", "false");
+        ret.set("util.debugging", "false");
+        ret.set("output.mode", "wasm");
+        ret.set("run.config.path", CHISEL_DEFAULT_CONFIG_PATH);
```
axic: I know this is const so it doesn't make a difference, but since the other defaults are here it may make sense just having the value here too.

chisel/src/result.rs
jakelang: Looks kinda like I spammed every letter in the word "Result" and then managed to finally type it correctly.

chisel/src/main.rs
jakelang: Looks strange when you run `chisel --help`

chisel/src/result.rs
```diff
@@ -0,0 +1,223 @@
...
+    pub fn results_mut(&mut self) -> &mut Vec<ModuleResult> {
+        &mut self.results
+    }
+
+    pub fn set_output_path(&mut self, path: PathBuf) {
+        self.output_path = path;
+    }
+
+    pub fn set_output_module(&mut self, module: &Module) {
+        self.output_module = Some(module.clone());
```
axic: I'd just take `Module` as value and not clone here. Leave cloning to the caller if it needs to be.

chisel/src/driver.rs
```diff
@@ -0,0 +1,544 @@
...
+    /// A canonicalized path was generated unsuccessfully. Left-hand is the config object name,
+    /// right-hand is the invalid path.
+    PathResolution(String, String),
+    /// An internal error occurred. Field 0 is the config object, during the execution of which the error occurred.
+    /// Field 1 is an additional informational message. Field 2 is the error generated.
+    Internal(String, String, Box<dyn Error>),
+}
+
+impl<'a> ChiselDriver<'a> {
+    pub fn new(config: &'a mut ChiselConfig) -> Self {
```
axic: The documentation above says config consumed here, so why not pass by value?

chisel/src/driver.rs
```diff
@@ -0,0 +1,544 @@
...
+    fn fire_internal(&mut self) {
+        let mut results = match &mut self.state {
+            DriverState::Ready => ChiselResult::new(),
+            DriverState::Error(_, r) => r.clone(),
+            DriverState::Done(_) => panic!("fire_internal() called on a completed driver"),
+        };
+
+        // Consume the rulesets in the configuration and execute each one.
+        while let Some((name, mut ruleset)) = self.config.rulesets_mut().pop_front() {
+            let mut rs_result = RSResult::new(name.clone());
```
axic: Yeah really want to get this "rs" stuff renamed to ruleset throughout, because it is easier to understand that way.

chisel/src/driver.rs
```diff
@@ -0,0 +1,544 @@
...
+
+            // Load binary.
+            chisel_debug!(1, "Running ruleset {}", name);
+            chisel_debug!(1, "Looking for binary path...");
+            let binary_path = if let Some(binary_path) = ruleset.options().get(&"file".to_string())
+            {
```
axic: rustfmt fail, this looks terrible.

chisel/src/driver.rs
```diff
@@ -0,0 +1,544 @@
...
+                    );
+                    return;
+                }
+            };
+
+            // Consume modules in ruleset and execute.
+            while let Some((name, module)) = ruleset.modules_mut().pop_front() {
+                chisel_debug!(1, "Executing module {}", &name);
+
+                // TODO: homogeneous module handling
```
axic: I think the solution to this giant mess is passing a `HashMap<String, String>` as configuration as part of `ChiselModule` or `ModulePreset`.

axic: Attempted it in #161.

chisel/src/result.rs
jakelang: The driver needs the module, so we have to pass it by reference.
I'll make some changes in the driver so this isn't necessary.

chisel/src/result.rs
```diff
@@ -93,7 +97,7 @@ impl RulesetResult {
     /// Write output module to specified file if the module was mutated.
     /// Returns Ok(false) if there is no mutation.
     /// Returns error on writer error or invalid mode.
-    pub fn write(&mut self, mode: &str) -> Result<bool, Box<dyn Error>> {
+    pub fn write(mut self, mode: &str) -> Result<bool, Box<dyn Error>> {
```
axic: Aha! :)

chisel/src/main.rs
```diff
@@ -7,360 +7,110 @@ extern crate serde_yaml;
...
-    fn get_modules(&self) -> &Vec<ModuleContext> {
-        &self.modules
-    }
-}
-
-impl ModuleContext {
-    fn fields(&self) -> (&String, &String) {
-        (&self.module_name, &self.preset)
-    }
+fn fail(code: i32, message: &str) -> ! {
```
axic: Looking at all the users of `fail`, all of them use `code = 1`. Do we need the `code` right now?

jakelang:
Split into different PRs and merged.

PR #97, opened 2019-05-31
Fix main signature for memory deployer

s1na:
This was the wat generated for the memory deployer:

```
(module
   (type (;0;) (func (param i32 i32)))
   (type (;1;) (func))
   (import "ethereum" "finish" (func (;0;) (type 0)))
   (func (;1;) (type 0) (param i32 i32))
   (func (;2;) (type 1)
     i32.const 0
     i32.const 519686
     call 0)
   (memory (;0;) 8)
   (export "main" (func 1))
   (export "memory" (memory 0))
...
```

However `(func 1)` has the signature for `ethereum::finish`, which resulted in invalid signature for `main`. (the `(func 1)` seems redundant to me anyway)


axic:
Rebased.

axic:
Great find!

PR #98, opened 2019-06-02
add remapexports

jwasinger:
Default preset for "ewasm" supports translating C wasm contracts which, when built with LLVM export the main function as "_main".

Does this look good?  If so, I'll add some unit tests.

### DIFF
--- a/chisel/src/main.rs
+++ b/chisel/src/main.rs
@@ -10,7 +10,7 @@ use std::fs::{read, read_to_string};
 use std::process;
 
 use libchisel::{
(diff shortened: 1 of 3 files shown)


jwasinger:
@axic suggested an alternative approach:

> Alex (axic), [02.06.19 01:35]
in order to properly operate, you'd want to rename the start to main and remove the startfunc entry in wasm
Alex (axic), [02.06.19 01:35]
i think this is a worthwhile new transformation in chisel

axic:
This is tracked in #99.

PR #100, opened 2019-06-02
Introduce FromWat

axic:
Perhaps this has to be a helper instead?

Depends on #108.

### DIFF
--- a/libchisel/src/fromwat.rs
+++ b/libchisel/src/fromwat.rs
@@ -1,0 +1,94 @@
+use super::{ModuleCreator, ModuleError};
+use crate::utils::*;
+use parity_wasm::elements::{deserialize_buffer, Module};
+use wabt::Wat2Wasm;
+
+/// Struct on which ModuleCreator is implemented.
+pub struct FromWat<'a> {
+    filename: &'a str,
+}
+
+impl<'a> FromWat<'a> {
+    pub fn new(filename: &'a str) -> Result<Self, ()> {
+        Ok(FromWat { filename: filename })
+    }
+}
+
+fn load_file(filename: &str) -> Result<Vec<u8>, ModuleError> {
+    use std::fs::File;
+    use std::io::prelude::*;
+    let mut file = File::open(filename)?;
+    let mut buf = Vec::new();
+    file.read_to_end(&mut buf)?;
+    Ok(buf)
+}
+
+impl From<wabt::Error> for ModuleError {
+    fn from(error: wabt::Error) -> Self {
+        use std::error::Error;
+        ModuleError::Custom(error.description().to_string())
+    }
+}
+
+fn convert_wat(input: &[u8]) -> Result<Module, ModuleError> {
+    // TODO: turn on relocatable(true)?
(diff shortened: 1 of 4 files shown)


libchisel/src/fromwat.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub filename: &'a String,
+}
+
+impl<'a> FromWat<'a> {
+    pub fn new(filename: &'a String) -> Result<Self, ()> {
+        Ok(FromWat { filename: filename })
+    }
+}
+
+fn load_file(filename: &str) -> Vec<u8> {
```
jakelang: Should return a `Result`, or at least use `expect` on the I/O calls.

libchisel/src/fromwat.rs
```diff
@@ -0,0 +1,96 @@
+extern crate wabt;
```
jakelang: No need for this, as we are using 2018 edition.

libchisel/src/fromwat.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+fn convert_wat(input: &[u8]) -> Result<Module, ModuleError> {
+    // TODO: turn on relocatable(true)?
+    let module = Wat2Wasm::new()
+        .canonicalize_lebs(true)
+        .write_debug_names(true)
+        .convert(&input)?;
+
+    let module = deserialize_buffer::<Module>(module.as_ref())?;
```
jakelang: We can clean this up by doing
```
deserialize_buffer::<Module>(module.as_ref())?
```
as the return, rather than explicitly binding it to the `module` ident and rewrapping it with `Ok`.

libchisel/src/fromwat.rs
```diff
@@ -0,0 +1,96 @@
+extern crate wabt;
+
+use super::{ModuleCreator, ModuleError};
+use crate::utils::*;
+use parity_wasm::elements::{deserialize_buffer, Module};
+use wabt::Wat2Wasm;
+
+/// Struct on which ModuleCreator is implemented.
+pub struct FromWat<'a> {
+    pub filename: &'a String,
```
jakelang: In my opinion we don't really need `&String`. It is more flexible with `&'a str` because `&String` would force it to be an owned string, whereas `String` already can be borrowed as a string slice.

jakelang: Also, it is a bit more idiomatic to expose getter methods rather than having public members. But it is not super important in this context.

axic: Well if the input to `new()` is not a `str` than this wont work.

jakelang: `String` can be borrowed as `&str`

jakelang: https://doc.rust-lang.org/std/string/struct.String.html#impl-AsRef%3Cstr%3E

axic: > Also, it is a bit more idiomatic to expose getter methods rather than having public members. But it is not super important in this context.

Actually it doesn't need to be `pub`.

PR #102, opened 2019-06-03
Add some useful (internal) helpers

axic:


libchisel/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+pub trait SerializationHelpers {
+    /// Deserialize bytecode to a Module.
+    fn from_slice(input: &[u8]) -> Module;
+    /// Serialize Module to bytecode. Serialization consumes the input.
+    fn to_vec(self) -> Vec<u8>;
+}
+
+impl SerializationHelpers for Module {
+    fn from_slice(input: &[u8]) -> Self {
+        deserialize_buffer::<Module>(&input).expect("valid Wasm bytecode")
```
jakelang: On failure, should this not be `invalid Wasm bytecode`? same for below.

libchisel/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+
+pub trait HasNamesSection {
+    /// Returns true if the module has a NamesSection.
+    fn has_names_section(&self) -> bool;
+}
+
+impl HasNamesSection for Module {
+    fn has_names_section(&self) -> bool {
+        // Lets forcefully parse in case it wasn't yet.
+        let mut module = self.clone().parse_names().unwrap();
```
jakelang: I would prefer an `expect()` here rather than unwrap.

libchisel/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+        module.names_section().is_some()
+    }
+}
+
+pub trait SerializationHelpers {
+    /// Deserialize bytecode to a Module.
+    fn from_slice(input: &[u8]) -> Module;
```
jakelang: Are you sure we would rather panic internally rather than return a `Result<T>`?

libchisel/src/utils.rs
```diff
@@ -0,0 +1,33 @@
...
+    /// Serialize Module to bytecode. Serialization consumes the input.
+    fn to_vec(self) -> Vec<u8>;
```
jakelang: Same as above.

libchisel/src/utils.rs
axic: Dunno, "expect valid wasm bytecode" sounded nice.

libchisel/src/utils.rs
axic: Yeah fair, forgot to change it amidst copy pastes.

libchisel/src/utils.rs
axic: That was considered as an option, but this right now is only using in testing code, so its fine.

libchisel/src/utils.rs
jakelang: I think it makes sense to do "Invalid Wasm bytecode/module" as the panic message does not say "Expected x". may cause confusion.

libchisel/src/utils.rs
jakelang: ok

Issue #99, opened 2019-06-02
Translator for start function (remapstart)


axic:
1. Rename the function pointed to by the start section to `main`. If there is another function named `main`, rename it (prefix it with `_`) to avoid conflict – if there this conflicts with another function, keep adding underscore prefixes until it won't.

2. Remove the start section. (See `dropstartfunction`.)

jwasinger:
Gonna take this one on

jwasinger:
> (See dropstartfunction.)

What do you mean by this?

axic:
> > (See dropstartfunction.)
>
> What do you mean by this? Referring to the pwasm API I am guessing.

https://github.com/wasmx/wasm-chisel/blob/master/libchisel/src/trimstartfunc.rs

jakelang:
@axic I see no real reason to drop the names section, as export names are unique. The names section exists solely as a custom section to aid debuggers and other binary tools.

axic:
The reason is that we were lazy to check if the exported names are duplicated in the names section.

jakelang:
They aren't, and the names section doesn't seem to matter anyway. Maybe introducing another module for that later would make more sense (as opposed to implicit behavior in `remapstart`).

axic:
We have another which removes sections. It does matter because we don't want to generate an invalid output.

This is the spec: https://webassembly.github.io/spec/core/appendix/custom.html#name-section

axic:
Agreed that other transformations changing the export section don't deal with the names section. We should clarify this and fix if needed, but we can do it after remapstart.

PR #107, opened 2019-06-04
Add support for remapstart

jwasinger:
Fixes https://github.com/wasmx/wasm-chisel/issues/99

### DIFF
--- a/libchisel/src/remapstart.rs
+++ b/libchisel/src/remapstart.rs
@@ -1,0 +1,304 @@
+use parity_wasm::elements::*;
+
+use super::{ModuleError, ModulePreset, ModuleTranslator};
+
+pub struct RemapStart;
+
+impl ModulePreset for RemapStart {
+    fn with_preset(preset: &str) -> Result<Self, ()> {
+        match preset {
+            // TODO refactor this later
+            "ewasm" => Ok(RemapStart {}),
+            _ => Err(()),
+        }
+    }
+}
+
+impl ModuleTranslator for RemapStart {
+    fn translate_inplace(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        Ok(remap_start(module))
+    }
+
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError> {
+        let mut ret = module.clone();
+        if remap_start(&mut ret) {
+            Ok(Some(ret))
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+// NOTE: This seems to be exported properly in later versions of parity-wasm.
+// TODO: When updated, use the proper method instead.
+fn section_order(s: &Section) -> u8 {
+    match s {
+        Section::Custom(_) => 0x00,
+        Section::Unparsed { .. } => 0x00,
+        Section::Type(_) => 0x1,
+        Section::Import(_) => 0x2,
+        Section::Function(_) => 0x3,
+        Section::Table(_) => 0x4,
+        Section::Memory(_) => 0x5,
+        Section::Global(_) => 0x6,
+        Section::Export(_) => 0x7,
+        Section::Start(_) => 0x8,
+        Section::Element(_) => 0x9,
+        Section::Code(_) => 0x0a,
+        Section::Data(_) => 0x0b,
+        Section::Name(_) => 0x00,
+        Section::Reloc(_) => 0x00,
+    }
+}
+
+/// Replace an exported function with another function, or export if unexported.
+fn remap_or_export_main(module: &mut Module, export_name: &str, func_idx: u32) {
+    let new_func_export = ExportEntry::new(export_name.to_string(), Internal::Function(func_idx));
+
+    if let Some(export_section) = module.export_section_mut() {
+        let export_section = export_section.entries_mut();
+        // If we find an export named `export_name`, replace it. Otherwise, append an entry to the
+        // section with the supplied func index.
+        if let Some(main_export_loc) = export_section
+            .iter_mut()
+            .position(|e| *e.field() == export_name.to_string())
+        {
+            export_section[main_export_loc] = new_func_export;
+        } else {
+            export_section.push(new_func_export);
+        }
+    } else {
+        let sections = module.sections_mut();
+        let new_export_section =
+            Section::Export(ExportSection::with_entries(vec![new_func_export]));
+
+        // If we can find a section that is supposed to be after exports, insert the new section at its position (shifts other elements to the right).
+        // Otherwise, append it at the end.
+        // NOTE: Assumes that the ordering of sections is otherwise correct.
+        if let Some(exports_loc) = sections.iter().position(|sec| section_order(&sec) > 0x7) {
+            sections.insert(exports_loc, new_export_section);
+        } else {
+            sections.push(new_export_section);
+        }
+    }
+}
+
+fn remap_start(module: &mut Module) -> bool {
+    if let Some(start_func_idx) = module.start_section() {
+        // Look for an export "main". If found, replace it with an export of the function to
+        // which the start section points.
+        remap_or_export_main(module, "main", start_func_idx);
+
+        // Remove the start section, leaving the "main" export as the entry point.
+        module.clear_start_section();
+
+        true
+    } else {
+        false
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::utils::SerializationHelpers;
+    use crate::{ModulePreset, ModuleTranslator};
+    use rustc_hex::FromHex;
+
+    #[test]
+    fn remapstart_mutation() {
+        //wat:
+        //(module
+        //    (import "env" "ethereum_useGas" (func (param i64)))
+        //    (memory 1)
+        //    (export "main" (func $main))
+        //    (export "memory" (memory 0))
+        //    (func $main2)
+        //    (func $main)
+        //    (start $main2)
+        //)
+
+        let wasm: Vec<u8> = FromHex::from_hex(
+            "0061736d0100000001080260017e0060
+000002170103656e760f657468657265756d5f75736547617300000303020101050301000107110
+2046d61696e0001066d656d6f727902000801020a070202000b02000b0020046e616d65010e0201
+046d61696e02056d61696e320209030001000001000200",
+        )
+        .unwrap();
+
+        let mut module = Module::from_slice(&wasm);
+        module = module.parse_names().unwrap();
+        assert!(module.names_section().is_some());
+        let start_idx = module
+            .start_section()
+            .expect("Module missing start function");
+
+        let new = RemapStart::with_preset("ewasm")
+            .unwrap()
+            .translate(&module)
+            .expect("Module internal error")
+            .expect("new module not returned");
+
+        assert!(
+            new.start_section().is_none(),
+            "start section wasn't removed"
+        );
+        assert!(new
+            .export_section()
+            .expect("Module missing export section")
+            .entries()
+            .iter()
+            .find(|e| e.field() == String::from("main")
+                && *e.internal() == Internal::Function(start_idx))
+            .is_some());
+    }
+
+    #[test]
+    fn remapstart_no_mutation() {
+        // (module
+        //    (import "env" "ethereum_useGas" (func (param i64)))
+        //    (memory 1)
+        //    (export "main" (func $main))
+        //    (export "memory" (memory 0))
+        //    (func $main)
+        //)
+
+        let wasm: Vec<u8> = FromHex::from_hex(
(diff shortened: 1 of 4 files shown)


libchisel/Cargo.toml
```diff
@@ -11,6 +11,6 @@ keywords = ["webassembly", "wasm", "blockchain", "ethereum"]
 edition = "2018"
 
 [dependencies]
-parity-wasm = "^0.35.6"
 rustc-hex = "1.0"
 byteorder = "1.2.4"
+parity-wasm = "^0.35.0"
```
axic: Why did you go back from 0.35.6 to 0.35.0? I think there was a reason requiring 0.35.6 as the minimum.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,228 @@
...
+                }
+            }
+        };
+
+    while !try_replace(module, &export_name, start_func_idx, export_main_func_idx) {
+        export_name = format!("_{}", export_name);
+    }
+}
+
+fn remap_start(module: &mut Module, translations: &RemapStart) -> bool {
```
axic: I don't think there is any need for this `translations` variable.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,228 @@
...
+    let mut ret = replace_main_export(module);
+
+    if ret {
+        let name_section_exists: bool =
```
axic: You can replace this entire blob with `module.names_section().is_some()`.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,228 @@
...
+            if let Some(NameSection::Function(_)) = module.names_section() {
+                true
+            } else {
+                false
+            };
+
+        if name_section_exists {
```
axic: You can replace this entire blob with `module.names_section().is_some()`.

axic: I still don't get why you need the names section to rename exported symbols. Those are not part of the names section.

jwasinger:
> check if a start section is present. If not, just quit.

Already implemented as described.

> check if an export named main exists. If yes, start (recursively) rename so that there is no main function.

I have a comment in the code because I wasn't sure whether to expect the module to be validly formed when `remapstart` is run.  But I see now that I will have to make a change to fail without panicking when main export is not found.

> insert a new export called main which points to the same function index what the start section was using

Already implemented as described.

> remove the start section

Already implemented as described.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        })
+        .expect("main export not found"); // TODO can I expect that the module is valid ewasm at this point?
+        */
+
+    let start_func_idx = module.start_section().unwrap();
```
jakelang: This can panic if the module doesn't have a start section. Please include `expect`, if not a `match` to handle the cases explicitly.

I know this is only called after the caller `remap_start` asserts that the start section is `Some`, but still harder to understand if the checking is done somewhere else in the code implicitly.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+                *export_entry = ExportEntry::new(
+                    export_name.to_string(),
+                    Internal::Function(replacement_func_idx),
+                );
+                ret = true;
+            }
+        }
+    }
+
+    return ret;
```
jakelang: `ret` will do

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+// if the main export doesn't refer to the same function as the start section,
+// point the main export towards the function referenced by the start section
+// remove the start section
+fn replace_export(module: &mut Module, export_name: &str, replacement_func_idx: u32) -> bool {
+    let mut ret = false;
+
+    /*
```
jakelang: Can you get rid of this big comment or is it going to be used?

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use super::{ModuleError, ModulePreset, ModuleTranslator};
+use parity_wasm::elements::*;
+
+pub struct RemapStart {}
+
+impl ModulePreset for RemapStart {
+    fn with_preset(preset: &str) -> Result<Self, ()> {
+        match preset {
+            "ewasm" => Ok(RemapStart {}),
```
jakelang: This is a bit strange but also my fault because the CLI config stuff is still pretty bad.
Maybe leave a note here to change later.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+impl ModuleTranslator for RemapStart {
+    fn translate_inplace(&self, module: &mut Module) -> Result<bool, ModuleError> {
+        Ok(remap_start(module))
+    }
+
+    fn translate(&self, module: &Module) -> Result<Option<Module>, ModuleError> {
+        let mut ret = module.clone();
+        let modified = remap_start(&mut ret);
+        if modified {
+            return Ok(Some(ret));
```
jakelang: can fold this into a cleaner expression without the c-style return statements like so:
```rust
if remap_start(&mut ret) {
    Ok(Some(ret))
} else {
    Ok(None)
}
```

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
+use std::collections::HashMap;
+
+use super::{ModuleError, ModulePreset, ModuleTranslator};
```
jakelang: Need to separate the `std`, external crate, and internal imports with one blank line.
*pedantry intensifies*

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn remapstart_no_mutation() {
+        // (module
+        //    (import "env" "ethereum_useGas" (func (param i64)))
+        //    (memory 1)
+        //    (export "main" (func $main))
+        //    (export "memory" (memory 0))
+        //    (func $main)
+        //)
+
+        let wasm: Vec<u8> = FromHex::from_hex("0061736d0100000001080260017e0060000002170103656e760f657468657265756d5f7573654761730000030201010503010001071102046d61696e0001066d656d6f727902000a040102000b").unwrap();
```
jakelang: Can you wrap this to the next lines so it fits in the screen width?

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+mod tests {
+    use super::*;
+    use crate::{ModulePreset, ModuleTranslator, ModuleValidator};
+    use parity_wasm;
+    use rustc_hex::FromHex;
+    use std::collections::HashMap;
+    use std::fs::File;
+    use std::io;
+    use std::io::prelude::*;
+
```
jakelang: Needs a test case for in-place translation

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+                            Some(*f)
+                        } else {
+                            None
+                        }
+                    }
+                    _ => None,
+                }) {
+                Some(main_export_idx) => main_export_idx,
+                None => {
+                    return false;
```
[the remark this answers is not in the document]
jwasinger: nope you can't.  that would be trying to assign `false` to `main_func_export_idx` instead of returning from the function early.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,181 @@
...
+                }
+            }
+        } else {
+            return false;
```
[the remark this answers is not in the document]
jwasinger: See above comment.

libchisel/src/remapstart.rs
jwasinger: removed.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,222 @@
+use parity_wasm::elements::*;
+use std::collections::HashMap;
+
+use super::{ModuleError, ModulePreset, ModuleTranslator};
+
+pub struct RemapStart {}
```
jakelang: you don't need the braces. you can simply put a `pub struct RemapStart;`

axic:
Lets simplify my last comment https://github.com/wasmx/wasm-chisel/pull/107#pullrequestreview-245932155:

As a way to organise I'd say it should:

1.  check if a start section is present. If not, just quit.
2. drop the names section (not sure if exports are duplicated in names, but lets just drop it and we can fix later)
3. check if an export named `main` exists. If yes, remove that entry from the export section.
4. insert a new export called `main` which points to the same function index what the start section was using
5. remove the start section

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,269 @@
...
+    if module.names_section().is_some() {
+        clear_names_section(module);
+        ret = true;
+    }
+
+    if replace_export_with_start(module, "main") {
+        ret = true;
+    }
+
+    if module.start_section().is_some() {
```
axic: This is not needed, we quit if there's no start section above.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,269 @@
...
+fn remap_start(module: &mut Module) -> bool {
+    let mut ret = false;
+
+    if module.start_section().is_none() {
+        return false;
+    }
+
+    if module.names_section().is_some() {
+        clear_names_section(module);
+        ret = true;
```
axic: Why there is a need to set these to true and not just return true in the end?

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,269 @@
...
+            break;
+        }
+    }
+    if rmidx < sections.len() {
+        sections.remove(rmidx);
+    }
+}
+
+// replace an exported function with the function pointed to by the start section
+fn replace_export_with_start(module: &mut Module, export_name: &str) -> bool {
```
axic: If this is a generic helper I'd rename it to `ensure_export_match_function_index(module: &mut Module, export_name: &str, function_idx: usize)` (usize or whatever the func index's type is)

axic:
@jwasinger can you squash this to a single commit?

libchisel/src/remapstart.rs
jwasinger: does `replace_export` work?

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,262 @@
...
+                *export_entry =
+                    ExportEntry::new(export_name.to_string(), Internal::Function(func_idx));
+                ret = true;
+            }
+        }
+    } else {
+        module
+            .sections_mut()
+            .push(Section::Export(ExportSection::with_entries(vec![
+                ExportEntry::new(export_name.to_string(), Internal::Function(func_idx)),
```
axic: Since all these functions need `export_name` to be a `String` I'd argue just change the function singature to use that or to have a `let export_name = export_name.to_string()` on the op.

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let wasm: Vec<u8> = FromHex::from_hex(
+            "0061736d0100000001080260017e0060
+000002170103656e760f657468657265756d5f75736547617300000303020101050301000107110
+2046d61696e0001066d656d6f727902000801020a070202000b02000b0020046e616d65010e0201
+046d61696e02056d61696e320209030001000001000200",
+        )
+        .unwrap();
+
+        let mut module: Module = parity_wasm::deserialize_buffer::<Module>(&wasm).unwrap();
```
axic: You can use the helpers from `utils` (see some other tests).

libchisel/src/remapstart.rs
```diff
@@ -0,0 +1,259 @@
...
+        } else {
+            Ok(None)
+        }
+    }
+}
+
+fn clear_names_section(module: &mut Module) {
+    let sections = module.sections_mut();
+    let mut rmidx = sections.len();
+    for (index, section) in sections.iter_mut().enumerate() {
```
axic: This can be pulled out into `find_names_section(module) -> Option(usize)` which then simplifies this code – no need for the `if` below.

libchisel/src/remapstart.rs
jakelang: I think it is not an issue here because the pwasm functions consume strings, so converting a `&str` to owned form inline is cleaner than instantiating a bunch of clones upfront to move into the functions/structs.

axic: Bunch of clones? Single variable.

jakelang: Yes but if you move a string you pass ownership. it is necessary to make copies for pwasm to consume because it does not accept by reference

PR #108, opened 2019-06-04
Better error handling in transformation modules

axic:
Follow up of #105 and #106.

### DIFF
--- a/libchisel/src/deployer.rs
+++ b/libchisel/src/deployer.rs
@@ -1,4 +1,5 @@
 use super::{ModuleCreator, ModuleError};
+use crate::utils::*;
 use parity_wasm::builder;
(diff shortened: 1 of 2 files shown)


libchisel/src/deployer.rs
```diff
@@ -65,27 +66,26 @@ fn deployer_code() -> Vec<u8> {
 }
 
 /// Returns a module which contains the deployable bytecode as a custom section.
-fn create_custom_deployer(payload: &[u8]) -> Module {
+fn create_custom_deployer(payload: &[u8]) -> Result<Module, ModuleError> {
```
jakelang: If we start to see `Result<Module, ModuleError>` everywhere we should include a `type Result<T> = ::std::result::Result<T, ModuleError>` in `lib.rs`

PR #115, opened 2019-06-11
Define ChiselModule trait

jakelang:
General purpose interface for chisel modules. Needed for CLI refactor.

### DIFF
--- a/libchisel/src/checkstartfunc.rs
+++ b/libchisel/src/checkstartfunc.rs
@@ -1,4 +1,4 @@
-use super::{ModuleError, ModuleValidator};
+use super::{ChiselModule, ModuleError, ModuleValidator};
 use parity_wasm::elements::Module;
 
 /// Struct on which ModuleValidator is implemented.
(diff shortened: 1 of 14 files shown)


libchisel/src/checkstartfunc.rs
```diff
@@ -84,4 +96,23 @@ mod tests {
...
+            0x03, 0x02, 0x01, 0x00, 0x07, 0x08, 0x01, 0x04, 0x6d, 0x61, 0x69, 0x6e, 0x00, 0x00,
+            0x0a, 0x04, 0x01, 0x02, 0x00, 0x0b,
+        ];
+
+        let module = deserialize_buffer::<Module>(&wasm).unwrap();
+
+        let checker = CheckStartFunc::new(false);
+        assert_eq!("checkstartfunc", checker.id());
+
+        let as_trait: &dyn ModuleValidator = checker.as_abstract();
```
axic: If you don't add these tests to the other modules, then I suggest to include this test in `lib.rs` next to the other basic module tests.

PR #121, opened 2019-06-25
Make DepGraphManager and Edge public, prune unused imports in depgraph.rs

jakelang:
Avoids private trait bound on public trait warning, which will be made an error in future rustc releases.

### DIFF
--- a/libchisel/src/depgraph.rs
+++ b/libchisel/src/depgraph.rs
@@ -1,6 +1,6 @@
 use std::collections::HashSet;
 
-use parity_wasm::elements::{CodeSection, Func, FuncBody, FunctionSection, Instruction, Module};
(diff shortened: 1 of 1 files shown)


libchisel/src/depgraph.rs
```diff
@@ -1,22 +1,22 @@
...
+pub struct Edge(u32, u32);
 
 /// Container struct for the function dependency graph.
 pub struct DepGraph {
     edges: HashSet<Edge>,
 }
 
 /// Private interface for managing the function dependency graph
-trait DepGraphManager {
+pub trait DepGraphManager {
```
axic: In general I'm not happy that we're exposing these internal functions on the public trait.

PR #129, opened 2019-07-23
CLI: Do not propagate fatal errors when constructing configuration

jakelang:
Depends: #128

### DIFF
--- a/chisel/src/main.rs
+++ b/chisel/src/main.rs
@@ -21,10 +21,7 @@ use serde_yaml::Value;
 
 // Error messages
 static ERR_NO_SUBCOMMAND: &'static str = "No subcommand provided.";
-static ERR_FAILED_OPEN_CONFIG: &'static str = "Failed to open configuration file.";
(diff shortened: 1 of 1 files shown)


chisel/src/main.rs
```diff
@@ -97,7 +96,7 @@ impl ChiselContext {
                 let mut config_itr = config_clone.iter();
                 // Read modules while there are still modules left.
                 while let Some(module) = config_itr.next() {
-                    module_confs.push(ModuleContext::from_yaml(module)?);
+                    module_confs.push(ModuleContext::from_yaml(module).expect("**placeholder**"));
```
axic: Placeholder?

axic:
Rebased.

chisel/src/main.rs
jakelang: Fixed in the next PR.

axic: Fixed it here with using `unwrap`.

PR #134, opened 2019-08-02
Add module kind method to ChiselModule

jakelang:


libchisel/src/lib.rs
```diff
@@ -212,7 +225,12 @@ mod tests {
         let validator = SampleModule {};
         assert_eq!(validator.id(), "Sample");
 
-        let as_trait = validator.as_abstract();
+        let opaque: &dyn ChiselModule<ObjectReference = &dyn ModuleValidator> =
+            &validator as &dyn ChiselModule<ObjectReference = &dyn ModuleValidator>;
```
axic: These 2 lines look really nasty.

jakelang: They can be removed in normal code. This is just super explicit so that I know what type is being produced from the expression.

libchisel/src/lib.rs
```diff
@@ -34,11 +41,13 @@ pub enum ModuleError {
 }
 
 /// Utility interface for chisel modules.
-pub trait ChiselModule<'a>: Sized {
+pub trait ChiselModule<'a> {
```
axic: Any reason to remove sized?

jakelang: Not needed, and stops me from casting between traits.

PR #140, opened 2019-08-23
DropSection::CustomSectionByName to take a string by value

jakelang:


libchisel/src/dropsection.rs
```diff
@@ -3,17 +3,17 @@ use super::{ChiselModule, ModuleError, ModuleKind, ModuleTranslator};
 use parity_wasm::elements::*;
 
 /// Enum on which ModuleTranslator is implemented.
-pub enum DropSection<'a> {
+pub enum DropSection {
     NamesSection,
     /// Name of the custom section.
-    CustomSectionByName(&'a String),
+    CustomSectionByName(String),
```
axic: Don't know why this done this way.

axic:
Looks good, but please first fix the CI.

libchisel/src/dropsection.rs
jakelang: A helper to construct DropSection which generates a string cannot return DropSection to caller if it contains a reference to a string whose scope is within the helper function.
There are workarounds but they aren't very ergonomic.

axic: Fixed my typo: Don't know why it was done with a reference initially.

PR #141, opened 2019-08-23
Update rust image version to Debian Buster

jakelang:


.circleci/config.yml
```diff
@@ -34,9 +34,12 @@ jobs:
           name: Build
           command: |
              cargo build --release
+             cargo clean
              cargo build --release --features wabt
+             cargo clean
```
axic: There's no need here to clean. (The tests build in debug mode.)

.circleci/config.yml
```diff
@@ -12,10 +12,10 @@ jobs:
             gcc --version
             g++ --version
             apt-get -y update
-            apt-get -y install libclang-3.8-dev clang-3.8 libclang-dev libgcc-6-dev libclang-common-3.8-dev cmake
+            apt-get -y install libclang-dev clang libclang-7-dev libgcc-8-dev libclang-common-7-dev cmake
```
axic: Why clang/llvm 7 and not 8?

jakelang: debian buster doesn't have these in the repo. We can add llvm-toolchain-8 manually.

PR #142, opened 2019-08-23
Clear compiler warnings

jakelang:


libchisel/src/remapimports.rs
```diff
@@ -124,6 +124,8 @@ impl<'a> ImportInterface<'a> {
     }
 }
 
+// NOTE: 'new()' is currently unused in the library but useful in the future.
+#[allow(dead_code)]
```
axic: Just add this on `new`.

PR #143, opened 2019-08-23
Properly declare wabt as a feature in CLI

jakelang:
NOTE: it is called wabt-utils because a feature cannot be named the same as a dependency


axic:
> it is called wabt-utils because a feature cannot be named the same as a dependency

That was an actual feature of cargo that feature names can match deps.

PR #144, opened 2019-08-26
Make DropSection treat the name section as a custom section

jakelang:
According to the spec, a name section is just a custom section with the name `"name"`. Although dropsection already has a mode to drop the names section, we should still treat the names section as a custom section for spec compliance.

### DIFF
--- a/libchisel/src/dropsection.rs
+++ b/libchisel/src/dropsection.rs
@@ -82,7 +82,9 @@ impl<'a> ModuleTranslator for DropSection {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::utils::SerializationHelpers;
     use parity_wasm::builder;
+    use rustc_hex::FromHex;
 
     #[test]
     fn keep_intact() {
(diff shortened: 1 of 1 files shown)


libchisel/src/dropsection.rs
```diff
@@ -44,6 +44,7 @@ fn names_section_index_for(module: &Module) -> Option<usize> {
 fn custom_section_index_for(module: &Module, name: &str) -> Option<usize> {
     module.sections().iter().position(|e| match e {
         Section::Custom(_section) => _section.name() == name,
+        Section::Name(_) => name == "name", // A name section is a custom section, but parity-wasm makes a distinction.
```
axic: The reasons are explained here: https://github.com/wasmx/wasm-chisel/blob/master/libchisel/src/utils.rs#L21

axic:
The real simplification here could be changing `find_index` to call `custom_section_index_for("name")` for the `NamesSection` case, adding the above change and removing `names_section_index_for`.

axic:
Just realised we don't have tests for the names section.

axic:
Could have added a test checking the behaviour of dropping the names section..

PR #146, opened 2019-08-27
Add test cases for dropping name section

jakelang:


libchisel/src/dropsection.rs
```diff
@@ -201,4 +201,44 @@ mod tests {
...
+            61696e",
+        )
+        .unwrap();
+
+        let mut module = Module::from_slice(&input)
+            .parse_names()
+            .expect("Should not fail");
+        assert!(custom_section_index_for(&module, "name").is_some());
+
+        let dropper = DropSection::CustomSectionByName("name".to_string());
```
axic: Can you do a clone of module and use both dropsection options for the namessection?

PR #150, opened 2019-08-28
Export module_from_slice() and Module::to_vec/from_slice() helpers

axic:


axic:
Merged this because using it in sentinel-rs. This will be changed in subsequent PRs to actually return errors properly.

PR #153, opened 2019-08-29
CLI: Introduce logging macro and global log level

jakelang:
Pulled from the CLI rewrite.
Also writes all user-facing messages to stderr, as primary program output goes in stdout (we don't have any as output is written to a file).

NOTE: Yes we are writing debug messages to stderr. The new CLI does this because unix mode will write wasm binaries to stdout unless otherwise specified. Does it matter in this PR? No, it will be overwritten by the other PR anyway.

### DIFF
--- a/chisel/src/logger.rs
+++ b/chisel/src/logger.rs
@@ -1,0 +1,31 @@
+static mut LOG_LEVEL: i32 = 0;
+
+#[macro_export]
+macro_rules! chisel_debug {
+    ($lvl:expr, $($arg:tt)*) => {
+        crate::logger::Logger::with_global_level().log($lvl, &format!($($arg)*));
+    }
+}
+
+/// Simple logging utility struct.
+pub struct Logger(i32);
+
+impl Logger {
+    pub fn with_global_level() -> Self {
+        unsafe { Logger(LOG_LEVEL) }
+    }
+
+    pub fn log<T: AsRef<str>>(&self, level: i32, message: T) {
+        if self.0 >= level {
+            eprintln!("{}", message.as_ref());
+        }
+    }
+}
+
+/// Set the global log level.
+// NOTE: Unsafe in a multithreaded context. Add mutex later when this is moved into the library.
+pub fn set_global_log_level(lvl: i32) {
+    unsafe {
+        LOG_LEVEL = lvl;
+    }
+}
--- a/chisel/src/main.rs
+++ b/chisel/src/main.rs
@@ -6,6 +6,8 @@ extern crate serde;
 extern crate serde_derive;
 extern crate serde_yaml;
 
+mod logger;
+
 use std::fs::{read, read_to_string};
 use std::process;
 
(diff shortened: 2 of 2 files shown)


chisel/src/logger.rs
```diff
@@ -0,0 +1,30 @@
...
+        crate::logger::Logger::with_global_level().log($lvl, &format!($($arg)*));
+    }
+}
+
+/// Simple logging utility struct.
+pub struct Logger(i32);
+
+impl Logger {
+    pub fn with_global_level() -> Self {
+        unsafe { Logger(LOG_LEVEL) }
```
axic: Does this needs any synchronisation or something since it accesses a global?

axic: Or at the bare minimum a big fat warning stating that this could use a mutex, but the CLI is structured in a way it should not.

chisel/src/logger.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    pub fn log<T: AsRef<str>>(&self, level: i32, message: T) {
+        if self.0 >= level {
+            eprintln!("{}", message.as_ref());
```
axic: Why is this going to stderr?

chisel/src/main.rs
```diff
@@ -280,7 +283,7 @@ fn execute_module(context: &ModuleContext, module: &mut Module) -> bool {
     } else {
         ret.unwrap_err()
     };
-    println!("\t{}: {}", name, module_status_msg);
+    eprintln!("\t{}: {}", name, module_status_msg);
```
axic: I think this should not be an error.

chisel/src/main.rs
```diff
@@ -302,7 +305,7 @@ fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
             let mut module = module.parse_names().expect("Failed to parse NamesSection");
 
             let original = module.clone();
-            println!("Ruleset {}:", context.name());
+            eprintln!("Ruleset {}:", context.name());
```
axic: I think this should not be an error.

chisel/src/logger.rs
jakelang: When we move things into the library I will, but for now we can safely assume that the CLI is all single thread. Will label it.

chisel/src/main.rs
```diff
@@ -150,7 +152,7 @@ impl ModuleContext {
 }
 
 fn err_exit(msg: &str) -> ! {
-    println!("{}: {}", crate_name!(), msg);
+    eprintln!("{}: {}", crate_name!(), msg);
```
axic: This is an error so `eprintln!` makes sense.

chisel/src/logger.rs
jakelang: Because it is a debug message.

axic: I guess its debatable, but most tools which do not output consumable data onto stdout, will use stdout for debug.

PR #157, opened 2019-09-08
Update parity-wasm to 0.40.2

axic:
This removes the need for the local Module::from_slice/to_vec implementation.


axic:
Apparently the `binaryen` test on CLI is still not working properly, because this didn't catch two missing changes in binaryenopt :sad:

PR #158, opened 2019-09-11
CLI: Introduce new config infrastructure

jakelang:


chisel/src/config.rs
```diff
@@ -0,0 +1,553 @@
...
+        let module_names: Vec<String> = modules
+            .split(',')
+            .filter_map(|c| match c {
+                "" => None, // If the user put a comma somewhere weird, we need to filter out the empty string.
+                any => Some(any.to_string()),
+            })
+            .collect();
+
+        // Filter out whitespace for options.
+        let options: String = options.chars().filter(|c| *c != ' ').collect();
```
axic: Hmm, cannot options contain spaces, such as `deployer.test="hello world"` ?

chisel/src/config.rs
```diff
@@ -0,0 +1,553 @@
...
+        // Iterate over each module and generate a config from its provided options.
+        for name in module_names.iter() {
+            // Iterate over the collected options and find the options corresponding to this
+            // module.
+            let relevant_options: Vec<String> = module_options
+                .iter()
+                .filter_map(|option| {
+                    // Split by dot here to isolate the module name.
+                    let option_pair: Vec<&str> = option.split('.').collect();
+                    assert!(option_pair.len() >= 1);
```
axic: It's `len = 1` if there was no dot, right?

jakelang: Yes, this assertion is there just to crash instead of doing something unpredictable with a zero-length vec. It should not be possible to reach, but in case something changes in the future I put it there.

chisel/src/config.rs
```diff
@@ -0,0 +1,553 @@
...
+
+                    // If the module name is the same as the one we're getting options for, then
+                    // yield the option.
+                    if *name == option_pair[0] {
+                        Some(option.to_string())
```
axic: Since its already split here, why not return a tuple?

chisel/src/config.rs
jakelang: The one case where I expect this to be an issue is with dropping a section by name. Even then, I think it is bad form for custom section names to contain spaces.

chisel/src/config.rs
jakelang: I thought it would be more readable to keep the collecting relevant options part isolated from the unmarshalling step.

chisel/src/config.rs
```diff
@@ -0,0 +1,553 @@
...
+    fn from_yaml(yaml: &Value) -> Result<Self, String>
+    where
+        Self: Sized,
+    {
+        let mut opts: HashMap<String, String> = HashMap::new();
+        let mut modules: VecDeque<(String, ModuleConfig)> = VecDeque::new();
+
+        let ruleset = yaml.as_mapping().expect("Already validated");
+        for (key, val) in ruleset.iter() {
+            match (key, val) {
```
axic: I think this match currently is not adding any benefit, unless the `String, any` case can be broken into `String, mapping` and `String, null` - otherwise we mix two stiles of programming to make it harder to read.

axic: But fine with it this way too. Not sure what's better.

chisel/src/config.rs
```diff
@@ -0,0 +1,553 @@
...
+        Self: Sized,
+    {
+        if yaml == &Value::Null {
+            Ok(ModuleConfig(HashMap::new()))
+        } else {
+            let opts = yaml.as_mapping().expect("Already validated");
+            let opts: HashMap<String, String> = opts
+                .iter()
+                .filter_map(|(key, val)| {
+                    // We ignore pairs which aren't string-string or string-bool at the moment.
```
axic: I think this comment is outdated.

jakelang: You're right. removing.

chisel/src/config.rs
jakelang: I think there isn't any specific logic which would warrant duplicating the code into two cases.

chisel/src/config.rs
```diff
@@ -0,0 +1,570 @@
...
+/// Left-hand value: name of ruleset
+/// Right-hand value: ruleset configuration
+#[derive(Debug)]
+pub struct ChiselConfig(VecDeque<(String, Ruleset)>);
+
+/// A single ruleset, containing options and a set of module configurations.
+#[derive(Debug)]
+pub struct Ruleset {
+    /// Top-level ruleset configurations, such as the input/output file paths.
+    opts: HashMap<String, String>,
```
axic: Nitpick: some parts use `opt`/`opts` other use `option`/`options`. I'd just go with the longer for readablity.

chisel/src/config.rs
```diff
@@ -0,0 +1,570 @@
...
+        Self: Sized,
+    {
+        // Filter out whitespace for names.
+        let modules: String = modules.chars().filter(|c| *c != ' ').collect();
+
+        // Collect module names.
+        let module_names: Vec<String> = modules
+            .split(',')
+            .filter_map(|c| match c {
+                "" => None, // If the user put a comma somewhere weird, we need to filter out the empty string.
```
axic: Shouldn't this be an error instead?

chisel/src/config.rs
```diff
@@ -0,0 +1,570 @@
...
+impl FromYaml for ModuleConfig {
+    fn from_yaml(yaml: &Value) -> Result<Self, String>
+    where
+        Self: Sized,
+    {
+        if yaml == &Value::Null {
+            Ok(ModuleConfig(HashMap::new()))
+        } else {
+            let opts = yaml.as_mapping().expect("Already validated");
+            let opts: HashMap<String, String> = opts
```
axic: If these two lines are merged is rustfmt breaking it up weirdly?

chisel/src/config.rs
```diff
@@ -0,0 +1,570 @@
...
+
+        let ruleset = yaml.as_mapping().expect("Already validated");
+        for (key, val) in ruleset.iter() {
+            match (key, val) {
+                // If string-string pair, treat it as an option.
+                (Value::String(k), Value::String(v)) => {
+                    opts.insert(k.to_string(), v.to_string());
+                }
+                // If string-map pair or a string-null pair (no special options), treat it as a module configuration.
+                (Value::String(k), v) => {
```
axic: For this to be similar to `impl FromYaml for ModuleConfig` I'd split this case into `(Value::String(k), Value::Mapping(v))` and `Value::Null`.

chisel/src/config.rs
jakelang: No, I just thought it would be more clear to assign `opts` in two expressions.

chisel/src/config.rs
jakelang: it's in a filter map expression, so it should be `Some` or `None`. This shouldn't be an error case tho.

chisel/src/config.rs
axic: This applies throughout the file :)

PR #159, opened 2019-09-19
Remove dependency on parity_wasm from chisel CLI

axic:
Follow up of #154. Also rolls back #150.

### DIFF
--- a/chisel/src/main.rs
+++ b/chisel/src/main.rs
@@ -309,7 +309,7 @@ fn execute_module(context: &ModuleContext, module: &mut Module) -> bool {
 
 fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
     if let Ok(buffer) = read(context.file()) {
-        if let Ok(module) = module_from_bytes(&buffer) {
+        if let Ok(module) = Module::from_bytes(&buffer) {
             // If we do not parse the NamesSection here, parity-wasm will drop it at serialisation
             // It is useful to have this for a number of optimisation passes, including binaryenopt and snip
             // TODO: better error handling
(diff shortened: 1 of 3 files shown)


chisel/src/main.rs
```diff
@@ -327,12 +325,13 @@ fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
 
             // If the module was mutated, serialize to file.
             if original != module {
+                let serialized = module.to_bytes().unwrap();
                 if let Some(path) = context.outfile() {
                     chisel_debug!(1, "Writing to file: {}", path);
-                    serialize_to_file(path, module).unwrap();
+                    write(path, serialized).unwrap();
```
jakelang: Can we use `expect` with a little description that the write operation failed for some reason?

chisel/src/main.rs
```diff
@@ -327,12 +325,13 @@ fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
 
             // If the module was mutated, serialize to file.
             if original != module {
+                let serialized = module.to_bytes().unwrap();
```
jakelang: Same here but for serialization failure.

libchisel/src/lib.rs
```diff
@@ -3,7 +3,7 @@ extern crate binaryen;
 extern crate parity_wasm;
 extern crate rustc_hex;
 
-use parity_wasm::elements::Module;
+pub use parity_wasm::elements::Module;
```
jakelang: Not needed yet, but perhaps we could benefit from a prelude of sorts containing re-exports and the like.

chisel/src/main.rs
axic: We could but your previous wasn't doing that either and I expected the big CLI PR will deal with it so wanted to minise the change here.

libchisel/src/lib.rs
axic: Though since this is now exported, we could remove `module_from_bytes`.

PR #160, opened 2019-09-19
Rename re-exported Module to WasmModule for clarity

axic:
Fixes #159.


libchisel/src/lib.rs
```diff
@@ -3,7 +3,7 @@ extern crate binaryen;
 extern crate parity_wasm;
 extern crate rustc_hex;
 
-pub use parity_wasm::elements::Module;
+use parity_wasm::elements::Module;
```
jakelang: Can do `pub use parity_wasm::elements::Module as WasmModule` if we also rename it in the rest of the codebase, which would have the side effect of making all the other code more readable.

PR #161, opened 2019-09-19
Add attempt for better module configuration

### LINKED ISSUE #135: Key-value configuration for modules
`ModulePreset` is extremely inflexible. Instead, modules should be configured in a more ad-hoc manner by way of a key-value mapping.

axic:
Closes #135.

### DIFF
--- a/libchisel/src/binaryenopt.rs
+++ b/libchisel/src/binaryenopt.rs
@@ -1,6 +1,8 @@
+use std::collections::HashMap;
+
 use parity_wasm::elements::Module;
 
-use super::{ChiselModule, ModuleError, ModuleKind, ModulePreset, ModuleTranslator};
+use super::{ChiselModule, ModuleConfig, ModuleError, ModuleKind, ModulePreset, ModuleTranslator};
 
 // FIXME: change level names
 pub enum BinaryenOptimiser {
@@ -26,6 +28,20 @@ impl<'a> ChiselModule<'a> for BinaryenOptimiser {
 
     fn as_abstract(&'a self) -> Self::ObjectReference {
         self as Self::ObjectReference
+    }
+}
+
+impl ModuleConfig for BinaryenOptimiser {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Ok(BinaryenOptimiser::O2)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        if let Some(preset) = config.get("preset") {
+            BinaryenOptimiser::with_preset(preset)
+        } else {
+            Err(ModuleError::NotSupported)
+        }
     }
 }
 
--- a/libchisel/src/checkstartfunc.rs
+++ b/libchisel/src/checkstartfunc.rs
@@ -1,6 +1,8 @@
+use std::collections::HashMap;
+
 use parity_wasm::elements::Module;
 
-use super::{ChiselModule, ModuleError, ModuleKind, ModuleValidator};
+use super::{ChiselModule, ModuleConfig, ModuleError, ModuleKind, ModuleValidator};
 
 /// Struct on which ModuleValidator is implemented.
 pub struct CheckStartFunc {
@@ -28,6 +30,23 @@ impl<'a> ChiselModule<'a> for CheckStartFunc {
 
     fn as_abstract(&'a self) -> Self::ObjectReference {
         self as Self::ObjectReference
+    }
+}
+
+impl ModuleConfig for CheckStartFunc {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        let require_start = if let Some(value) = config.get("require_start") {
+            value == "true"
+        } else {
+            false
+        };
+        Ok(CheckStartFunc {
+            start_required: require_start,
+        })
     }
 }
 
--- a/libchisel/src/deployer.rs
+++ b/libchisel/src/deployer.rs
@@ -1,7 +1,9 @@
+use std::collections::HashMap;
+
 use parity_wasm::builder;
 use parity_wasm::elements::{CustomSection, Module};
 
-use super::{ChiselModule, ModuleError, ModuleKind, ModulePreset, ModuleTranslator};
+use super::{ChiselModule, ModuleConfig, ModuleError, ModuleKind, ModulePreset, ModuleTranslator};
 
 /// Enum on which ModuleTranslator is implemented.
 pub enum Deployer {
@@ -22,6 +24,20 @@ impl<'a> ChiselModule<'a> for Deployer {
 
     fn as_abstract(&'a self) -> Self::ObjectReference {
         self as Self::ObjectReference
+    }
+}
+
+impl ModuleConfig for Deployer {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        if let Some(preset) = config.get("preset") {
+            Deployer::with_preset(preset)
+        } else {
+            Err(ModuleError::NotSupported)
+        }
     }
 }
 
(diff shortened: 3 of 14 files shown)


libchisel/src/checkstartfunc.rs
```diff
@@ -14,6 +14,25 @@ impl CheckStartFunc {
...
 }
 
+use std::collections::HashMap;
+
+impl ModuleConfig for CheckStartFunc {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
```
axic: Could also split this into two (and that would remove the need for default):
- "with config string"
- "with complex config"

e.g.
```
fn with_config_string(config: &str)
fn with_config_map(config: &HashMap<String, String>)
```

and then `with_config_string(&"")` would mean "defaults"

jakelang: I think key-val configuration should be fine for now. No need to further complicate things with the config string, and we can definitely just implement the config string as a specified key `config` or something in the hashmap.

libchisel/src/checkstartfunc.rs
```diff
@@ -14,6 +14,25 @@ impl CheckStartFunc {
     }
 }
 
+use std::collections::HashMap;
+
+impl ModuleConfig for CheckStartFunc {
+    fn with_defaults() -> Result<Self, ModuleError> {
```
jakelang: We could just implement this via `Default`, I think.

axic: I know, but the point is to require it for every module to avoid the need for more magic in `ChiselModule`.

libchisel/src/checkstartfunc.rs
axic: True, but the downside is the terrible syntax needed for creating a hashmap and passing it. So for simple modules a string is much more simple.

jakelang: Perhaps the config string method could just wrap that logic?

axic: What do you mean by wrap? Define a way to encode hashmaps?

libchisel/src/checkstartfunc.rs
axic: And this one has a `Result`.

libchisel/src/checkstartfunc.rs
jakelang: `with_config_string` would just call `with_config` where a dedicated key such as `config_string` is set to the value provided.

libchisel/src/checkstartfunc.rs
jakelang: Why would the default be able to fail?

jakelang: nvm

libchisel/src/binaryenopt.rs
```diff
@@ -29,6 +31,21 @@ impl<'a> ChiselModule<'a> for BinaryenOptimiser {
     }
 }
 
+impl ModuleConfig for BinaryenOptimiser {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Ok(BinaryenOptimiser::O0)
```
jakelang: Wouldn't it make more sense to do `-O2` or the highest level which doesn't sacrifice size for speed?

libchisel/src/deployer.rs
```diff
@@ -25,6 +25,23 @@ impl<'a> ChiselModule<'a> for Deployer {
...
+use std::collections::HashMap;
+
+impl ModuleConfig for Deployer {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        let preset = config.get("preset");
+        if preset.is_none() {
```
jakelang: This expression can be simplified:
```
if let Some(preset) = config.get("preset") {
    Deployer::with_preset(preset)
} else {
    Err(ModuleError::NotSupported)
}
```

libchisel/src/dropsection.rs
```diff
@@ -29,6 +32,34 @@ impl<'a> ChiselModule<'a> for DropSection {
     }
 }
 
+impl From<std::num::ParseIntError> for ModuleError {
```
jakelang: Could make this more generic, and support all kinds of errors:
```
impl From<E: Error> for ModuleError {
    fn from(error: E) -> Self {
        ModuleError::Custom(error.description().to_string())
    }
}

File: libchisel/src/dropsection.rs [hunk 6]
```diff
@@ -29,6 +32,34 @@ impl<'a> ChiselModule<'a> for DropSection {
...
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        for (key, val) in config.iter() {
+            return match key.as_str() {
+                "names" => Ok(DropSection::NamesSection),
+                "custom_by_name" => Ok(DropSection::CustomSectionByName(val.clone())),
+                "custom_by_index" => {
+                    Ok(DropSection::CustomSectionByIndex(str::parse::<usize>(val)?))
+                }
+                "unknown_by_index" => Ok(DropSection::UnknownSectionByIndex(str::parse::<usize>(
+                    val,
```
axic: rustfmt is rather weird sometimes

libchisel/src/dropsection.rs
```diff
@@ -29,6 +32,34 @@ impl<'a> ChiselModule<'a> for DropSection {
...
+    }
+}
+
+impl ModuleConfig for DropSection {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        for (key, val) in config.iter() {
```
jakelang: Why do we need the loop here if it will always return on first iteration?

libchisel/src/deployer.rs
axic: Yeah now it can, before we had #171 this wasn't as simple.

libchisel/src/remapimports.rs
```diff
@@ -29,6 +30,23 @@ impl<'a> ChiselModule<'a> for RemapImports<'a> {
...
+use std::collections::HashMap;
+
+impl<'a> ModuleConfig for RemapImports<'a> {
+    fn with_defaults() -> Result<Self, ModuleError> {
+        Err(ModuleError::NotSupported)
+    }
+
+    fn with_config(config: &HashMap<String, String>) -> Result<Self, ModuleError> {
+        let preset = config.get("preset");
+        if preset.is_none() {
```
jakelang: Same here: https://github.com/wasmx/wasm-chisel/pull/161/files#r328288401

libchisel/src/dropsection.rs
axic: Legacy. I guess it could pick a random key/val and do it from there.

libchisel/src/dropsection.rs
axic: True, though I was considering here to wrap it with some text or use `NotSupported`.

libchisel/src/dropsection.rs
jakelang: That would result in some strange behavior, I feel.
Perhaps enforcing mutual exclusion of dropsection modes here would make the best sense.

axic: Yeah I'm torn on that one. Enforcing that will add a ton of complexity – that's how I started first.

libchisel/src/lib.rs
```diff
@@ -80,6 +81,17 @@ pub trait ModulePreset {
         Self: std::marker::Sized;
 }
 
+// TODO: move this to be part of ChiselModule
```
jakelang: I'm undecided as to whether that is a good idea.

libchisel/src/dropsection.rs
axic: Apparently this is unstable:  https://github.com/rust-lang/rust/issues/52662

jakelang: You could have this blanket impl and then have specialized impls for specific traits

PR #162, opened 2019-09-20
Abstract configuration API (attempt 1)

jakelang:
NOTE: Only works on nightly but I can remove the hashmap as default.

### DIFF
--- a/libchisel/src/lib.rs
+++ b/libchisel/src/lib.rs
@@ -1,3 +1,5 @@
+#![feature(associated_type_defaults)]
+
 #[cfg(feature = "binaryen")]
 extern crate binaryen;
 extern crate parity_wasm;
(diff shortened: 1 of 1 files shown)


libchisel/src/lib.rs
```diff
@@ -54,6 +59,51 @@ pub trait ChiselModule<'a> {
...
+    K: Clone + Hash + Eq,
+    V: Clone,
+{
+    fn insert(&mut self, key: &K, value: &V) -> Option<V>;
+
+    fn get(&self, key: &K) -> Option<&V>;
+
+    fn get_mut(&mut self, key: &K) -> Option<&mut V>;
+
+    fn remove(&mut self, key: &K) -> Option<V>;
```
axic: Only `get` is needed for chisel.

jakelang: Yeah, I know. I was treating this as a general abstraction for keyvalue containers.

jakelang:
Outdated

PR #163, opened 2019-09-20
Introduce new CLI result handling

jakelang:


chisel/src/result.rs
```diff
@@ -0,0 +1,222 @@
...
+
+#[derive(Clone)]
+/// Main result structure returned by ChiselDriver, containing a manifest of modules executed and
+/// exposing methods to write output from translators and creators.
+pub struct ChiselResult(Vec<RulesetResult>);
+
+#[derive(Clone)]
+/// Ruleset result structure.
+pub struct RulesetResult {
+    rs_name: String,
```
axic: `ruleset_name`?

chisel/src/result.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    pub fn set_output_path(&mut self, path: PathBuf) {
+        self.output_path = path;
+    }
+
+    pub fn set_output_module(&mut self, module: Module) {
+        self.output_module = Some(module);
+    }
+
+    pub fn write_output_to_file(&self) -> Result<bool, Box<dyn Error>> {
```
axic: Since a lot of this is duplicated across the three `write_output`, maybe we could use an enum for "output kind" and a match? Perhaps that's too much, but at least deduplicating the serialization stuff would be useful.

chisel/src/result.rs
```diff
@@ -0,0 +1,222 @@
...
+        } else {
+            Ok(false)
+        }
+    }
+
+    pub fn write_output_to_file_hex(&self) -> Result<bool, Box<dyn Error>> {
+        if let Some(ref module) = self.output_module {
+            match module.clone().to_bytes() {
+                Ok(buf) => {
+                    let hex_encoded = encode(buf);
```
axic: Can you use `encode` here as `hex::encode` to make it more readable?

chisel/src/result.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.output_path = path;
+    }
+
+    pub fn set_output_module(&mut self, module: Module) {
+        self.output_module = Some(module);
+    }
+
+    /// Get a ChiselOutputWriter from the ruleset and consumes the output module.
+    pub fn get_writer(&mut self, mode: &str) -> Option<ChiselOutputWriter> {
+        if let Some(module) = self.output_module.take() {
```
axic: Since `take` replaced `output_module` with `None` it may as well be simpler taking `self` by value and not reference to avoid all this?

The confusing part is that calling `get_writer` successfully make any further call return empty output.

chisel/src/result.rs
```diff
@@ -0,0 +1,223 @@
...
+            Err(e) => Err(e.into()),
+        }
+    }
+}
+
+impl Display for ChiselResult {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.0
+            .iter()
+            .map(|rs_result| write!(f, "{}", rs_result))
```
axic: For consistency name this `ruleset_result`.

chisel/Cargo.toml
```diff
@@ -16,6 +16,8 @@ clap = "2.32.0"
 serde = "1.0.80"
 serde_derive = "1.0.80"
 serde_yaml = "0.8.7"
+hex = "0.4.0"
+ansi_term = "0.12.1"
 
 [features]
 default = []
```
axic:
```suggestion
default = ["wabt-utils"]
wabt-utils = ["wabt", "libchisel/wabt"]
```

chisel/Cargo.toml
```diff
@@ -16,6 +16,8 @@ clap = "2.32.0"
 serde = "1.0.80"
 serde_derive = "1.0.80"
 serde_yaml = "0.8.7"
+hex = "0.4.0"
+ansi_term = "0.12.1"
```
axic: Move these to the top.

chisel/Cargo.toml
```diff
@@ -11,13 +11,15 @@ keywords = ["webassembly", "wasm", "blockchain", "ethereum"]
...
 clap = "2.32.0"
+hex = "0.4.0"
+libchisel = { path = "../libchisel", version = "0.5.0" }
 serde = "1.0.80"
 serde_derive = "1.0.80"
 serde_yaml = "0.8.7"
 
 [features]
-default = []
+default = ["wabt"]
```
axic: Hm not sure about that, it is kind of slow to compile :(

chisel/src/result.rs
```diff
@@ -0,0 +1,255 @@
...
+
+use std::error::Error;
+use std::fmt::{self, Display};
+use std::fs::write;
+use std::path::{Path, PathBuf};
+
+use libchisel::wabt;
+use libchisel::{Module, ModuleError};
+
+use ansi_term::Colour::{Green, Red, Yellow};
```
axic: Lets enforce the ordering here then :wink:

libchisel/src/lib.rs
```diff
@@ -1,7 +1,9 @@
 #[cfg(feature = "binaryen")]
 extern crate binaryen;
-extern crate parity_wasm;
+pub extern crate parity_wasm;
```
axic: Why is this needed?

chisel/src/result.rs
```diff
@@ -0,0 +1,255 @@
...
+mod tests {
+    use libchisel::parity_wasm::builder;
+
+    use super::*;
+
+    #[test]
+    fn writer_deny_raw_binary_to_stdout() {
+        let mut ruleset_result = {
+            let mut result = RulesetResult::new("Test".to_string());
+            let module = builder::module().build();
```
axic: Can just use `Module::default()` which is already exported.

libchisel/src/lib.rs
jakelang: builder api needed for tests.

chisel/src/result.rs
```diff
@@ -0,0 +1,276 @@
...
+    }
+}
+
+impl ChiselOutputWriter {
+    /// Serializes and writes the contained module.
+    pub fn write(self) -> Result<(), Box<dyn Error>> {
+        let ret = match self {
+            ChiselOutputWriter::Bin(module, path) => {
+                if *path == PathBuf::from("/dev/stdout") || *path == PathBuf::from("/dev/stderr") {
+                    return Err("cannot write raw binary to a standard stream".into());
```
axic: Why is this disallowed?

jakelang: Shits out escape codes you might not want, kinda like reading from /dev/random

PR #164, opened 2019-09-24
Clear up use statements

axic:


libchisel/src/dropsection.rs
```diff
@@ -80,6 +80,7 @@ impl<'a> ModuleTranslator for DropSection {
 mod tests {
     use super::*;
```
jakelang: If we're following the rule of
```
std imports
extern imports
local imports
```
then this goes at the bottom, no?

libchisel/src/repack.rs
```diff
@@ -43,6 +44,7 @@ impl ModuleTranslator for Repack {
 mod tests {
     use super::*;
     use parity_wasm::builder;
+    use parity_wasm::elements::CustomSection;
```
jakelang: Same here

libchisel/src/dropsection.rs
axic: I didn't care about tests.

PR #165, opened 2019-09-25
Use new config infrastructure

axic:


chisel/src/main.rs
```diff
@@ -345,6 +255,49 @@ fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
...
+        } else {
+            config.unwrap()
+        }
+    } else {
+        err_exit("Failed to parse YAML configuration")
+    }
+}
+
+impl From<&(String, config::Ruleset)> for ChiselContext {
+    fn from(input: &(String, config::Ruleset)) -> ChiselContext {
```
jakelang: Do we want to take a reference or can we simply take it by value and consume it?

chisel/src/main.rs
```diff
@@ -345,6 +255,49 @@ fn chisel_execute(context: &ChiselContext) -> Result<bool, &'static str> {
...
+        let modules: Vec<ModuleContext> =
+            input.1.modules().iter().fold(Vec::new(), |mut acc, x| {
```
jakelang: Could use `m` or `module` in the closure for extra clarity.

chisel/src/main.rs
axic: That needs a change in the caller, this was the simplest approach.

axic: It is actually worse trying to do that:
```
320 |             let ctx: ChiselContext = (*ctx).into();
    |                                      ^^^^^^ move occurs because `*ctx` has type `(std::string::String, config::Ruleset)`, which does not implement the `Copy` trait
```

PR #167, opened 2019-09-25
Add new CLI option handling

jakelang:


chisel/src/options.rs
```diff
@@ -0,0 +1,118 @@
...
+                acc.push_str(&format!("{},", val));
+                acc
+            });
+            self.set("oneliner.modules.options", &values_collected);
+        }
+        if let Some(value) = matches.value_of("FILE") {
+            self.set("oneliner.file", value);
+        }
+        if let Some(value) = matches.value_of("OUTPUT_PATH") {
+            self.set("oneliner.out", value);
```
axic: This is called `output` in other places.

chisel/src/options.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        if let Some(value) = matches.value_of("OUTPUT_PATH") {
+            self.set("oneliner.output", value);
+        }
+        if let Some(value) = matches.value_of("OUTPUT_MODE") {
+            match value {
+                val @ "bin" | val @ "wat" | val @ "hex" => {
+                    self.set("output.mode", val);
+                }
+                _ => panic!("CLI parser only accepts 'bin', 'wat', or 'hex'"),
```
axic: Why not return an error here?

jakelang: This function does not return a set of flags, it applies arguments from `clap` to the flag set.

axic: Terrible. Is there a way to signal to clap what values are accepted?

axic: It does: https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.possible_values

jakelang: Yes, in the app declaration. It should be done there, which is why it makes sense to panic here. This simply should be unreachable.

axic: This panic is fine as long as the above `possible_values` will be used so that clap doesn't even let this pass through to the panic.

jakelang: exactly

PR #168, opened 2019-09-25
Introduce new CLI driver skeleton

jakelang:


chisel/src/driver.rs
```diff
@@ -0,0 +1,245 @@
...
+    Module,
+    ModulePreset,
+    ModuleTranslator,
+    ModuleValidator,
+};
+
+/// State machine implementing the main chisel execution loop. Consumes ChiselConfig and returns
+/// ChiselResult, with an intermediate state returned to allow error handling.
+pub struct ChiselDriver<'a> {
+    config: &'a mut ChiselConfig,
```
axic: The comment above says "it consumes the config". So why have it as a ref here?

chisel/src/driver.rs
```diff
@@ -0,0 +1,245 @@
...
+    fn fire_internal(&mut self) {
+        let mut results = match &mut self.state {
+            DriverState::Ready => ChiselResult::new(),
+            DriverState::Error(_, r) => r.clone(),
+            DriverState::Done(_) => panic!("fire_internal() called on a completed driver"),
+        };
+
+        // Consume the rulesets in the configuration and execute each one.
+        while let Some((name, mut ruleset)) = self.config.rulesets_mut().pop_front() {
+            let mut rs_result = RulesetResult::new(name.clone());
```
axic: `ruleset_result`?

chisel/src/driver.rs
```diff
@@ -0,0 +1,245 @@
...
+
+use std::error::Error;
+use std::fmt::{self, Display};
+use std::fs::{canonicalize, read};
+use std::path::PathBuf;
+
+use crate::config::ChiselConfig;
+use crate::result::{ChiselResult, ModuleResult, RulesetResult};
+
+use libchisel::{
```
axic: If its a skeleton tall this can be deleted.

chisel/src/driver.rs
```diff
@@ -0,0 +1,245 @@
...
+                panic!("fire() called on a completed driver");
+            }
+        }
+        &self.state
+    }
+
+    fn fire_internal(&mut self) {
+        let mut results = match &mut self.state {
+            DriverState::Ready => ChiselResult::new(),
+            DriverState::Error(_, r) => r.clone(),
```
axic: `r` -> `result` ?

chisel/src/driver.rs
```diff
@@ -0,0 +1,245 @@
...
+                );
+                return;
+            };
+
+            // Look for output path and set.
+            let output_path =
+                if let Some(output_path) = ruleset.options().get(&"output".to_string()) {
+                    chisel_debug!(1, "Found output path: {}", &output_path);
+                    PathBuf::from(output_path)
+                } else if let Some(output_path) = ruleset.options().get(&"out".to_string()) {
```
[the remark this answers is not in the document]
jakelang: So we're backwards-compatible with the old CLI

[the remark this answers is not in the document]
axic: The old one used `output`.

chisel/src/driver.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// A canonicalized path was generated unsuccessfully. Left-hand is the config object name,
+    /// right-hand is the invalid path.
+    PathResolution(String, String),
+    /// An internal error occurred. Field 0 is the config object, during the execution of which the error occurred.
+    /// Field 1 is an additional informational message. Field 2 is the error generated.
+    Internal(String, String, Box<dyn Error>),
+}
+
+impl ChiselDriver {
+    pub fn get_results(self) -> Result<ChiselResult, ()> {
```
axic: Hm, why not option is this isn't really outputting the error?

chisel/src/driver.rs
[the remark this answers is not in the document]
jakelang: Wait, I actually included this here to accept both.

chisel/src/driver.rs
```diff
@@ -0,0 +1,227 @@
...
+        match self.state {
+            DriverState::Ready => Err(()),
+            DriverState::Error(_, result) => Ok(result),
+            DriverState::Done(result) => Ok(result),
+        }
+    }
+
+    pub fn new(config: ChiselConfig) -> Self {
```
axic: Move this before `get_results`?

chisel/src/driver.rs
jakelang: Matter of choice, I consider calling `get_results()` on `DriverState::Ready` to be an error.

axic: Its confusing because why isn't it an error on the error state? :)

Since `fire()` is supposed to be used by the caller to figure out the state, this one here is only for consuming the final result. I'd suggest to rename it to `take_result` and make it return an option.

chisel/src/driver.rs
jakelang: Done

chisel/src/driver.rs
jakelang: > Why isn't it an error on the error state?
When `norecover` is enabled, it leaves the option open to take the result and write whatever did succeed.

chisel/src/driver.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    pub fn take_result(self) -> Result<ChiselResult, ()> {
+        match self.state {
+            DriverState::Ready => Err(()),
+            DriverState::Error(_, result) => Ok(result),
+            DriverState::Done(result) => Ok(result),
+        }
+    }
+
+    pub fn fire(&mut self) -> &DriverState {
```
axic: Is there actually a need for this wrapper? It seems the comments from here could be just moved into the match in `fire_internal`.

jakelang: I thought it would be more readable to isolate the guts from the initial matcher, but this was back when we had the huge match statement.

chisel/src/driver.rs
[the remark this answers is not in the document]
axic: It is `output` in the old (current) yaml.

chisel/src/driver.rs
axic: Ah that makes sense. Well hope we wont have that monstrosity.

chisel/src/driver.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    pub fn take_result(self) -> Result<ChiselResult, ()> {
+        match self.state {
+            DriverState::Ready => Err(()),
+            DriverState::Error(_, result) => Ok(result),
+            DriverState::Done(result) => Ok(result),
+        }
+    }
+
+    fn fire(&mut self) -> &DriverState {
```
axic: `pub`

chisel/src/driver.rs
```diff
@@ -0,0 +1,228 @@
...
+    use std::panic::catch_unwind;
+
+    use super::*;
+    use crate::config::{ChiselConfig, FromArgs};
+
+    #[test]
+    fn take_result_ready() {
+        let result = catch_unwind(|| {
+            let config = ChiselConfig::from_args("test", "test.foo=bar").expect("Cannot fail");
+            let driver = ChiselDriver::new(config);
```
axic: Can you move these two out so that unwind only catches the panic in `take_result`?

jakelang: No, it doesn't allow config and driver to cross unwind boundaries.

axic: Or add another test case with these two lines succeeding.

jakelang: Even with `unsafe`

jakelang: The success test requires fire().

PR #173, opened 2019-09-27
Driver: Implement module handling in a helper

jakelang:


chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
...
+        &mut self,
+        name: String,
+        module: ModuleConfig,
+        wasm: &mut Module,
+    ) -> Result<ModuleResult, DriverError> {
+        let result = match name.as_str() {
+            "checkfloat" => {
+                let checkfloat = CheckFloat::new();
+                let module_result = checkfloat.validate(wasm);
+                ModuleResult::Validator(name.clone(), module_result)
```
axic: Why not wrap these in `Ok()` and remove the match assignment?

chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
...
+                    ModuleResult::Validator(name.clone(), module_result)
+                } else {
+                    chisel_debug!(1, "checkstartfunc missing field 'require_start'");
+                    return Err(DriverError::MissingRequiredField(
+                        name.clone(),
+                        "require_start".to_string(),
+                    ));
+                }
+            }
+            "deployer" => unimplemented!("Creator modules are not supported yet."),
```
axic: This is not a creator anymore.

chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
...
+            "dropsection" => {
```
axic: Suggest this implementation is dropped as-is given its not there in the current CLI.

chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
...
+                    (None, _) => {
+                        chisel_debug!(1, "dropsection missing field 'mode'");
+                        return Err(DriverError::MissingRequiredField(
+                            name.clone(),
+                            "mode".to_string(),
+                        ));
+                    }
+                }
+            }
+            "fromwat" => unimplemented!("Creator modules are not supported yet."),
```
axic: Can be dropped for same reason.

chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
...
+                        ));
+                    }
+                } else {
+                    chisel_debug!(1, "verifyimports missing field 'preset'");
+                    return Err(DriverError::MissingRequiredField(
+                        name.clone(),
+                        "preset".to_string(),
+                    ));
+                }
+            }
```
axic: Please add `dropnames` and `binaryenopt` though for the moment.

chisel/src/driver.rs
```diff
@@ -184,6 +205,278 @@ impl ChiselDriver {
         self.state = DriverState::Done(results);
         &self.state
     }
+
+    pub fn execute_module(
+        &mut self,
+        name: String,
```
axic: Can't this be a `&str`?

axic: Ah ok, you're consuming `name` and `module` from the caller.

chisel/src/driver.rs
axic: Can't you just consume `name` here (and the rest of the places below)?

jakelang: It'd be more concise but I think it looks a bit strange to have an `Ok()` wrapping an expression which could actually short-circuit and return `Err()` mid-evaluation

chisel/src/driver.rs
jakelang: We have a module called `dropnames`?

axic: No, but we have a case in the current CLI called `dropnames`.

chisel/src/driver.rs
```diff
@@ -184,6 +195,235 @@ impl ChiselDriver {
...
+                } else {
+                    chisel_debug!(1, "binaryenopt missing field 'preset'");
+                    return Err(DriverError::MissingRequiredField(
+                        name,
+                        "preset".to_string(),
+                    ));
+                }
+            }
+            other => {
+                return Err(DriverError::ModuleNotFound(name.clone(), other.to_string()));
```
axic: No need for clone here?

chisel/src/driver.rs
```diff
@@ -184,6 +195,235 @@ impl ChiselDriver {
...
+                } else {
+                    chisel_debug!(1, "verifyimports missing field 'preset'");
+                    return Err(DriverError::MissingRequiredField(
+                        name,
+                        "preset".to_string(),
+                    ));
+                }
+            }
+            #[cfg(feature = "binaryen")]
+            "binaryenopt" => {
```
axic: Doesn't make any difference but this it out of lexicographic order (while the rest is in order).

chisel/src/driver.rs
jakelang: It is needed, since it's in a scope where name is borrowed. See line 206

chisel/src/driver.rs
jakelang: I thought it would make more sense to separate conditionally compiled cases.

chisel/src/driver.rs
jakelang: NVM, it is actually needed because we use `other.to_string()` as the second argument.

PR #176, opened 2019-10-25
Rework cmd_run to use new driver and config systems

jakelang:


chisel/src/main.rs
```diff
@@ -6,65 +6,102 @@ mod driver;
...
+        .arg(
+            Arg::with_name("MODULES")
+                .short("m")
+                .long("modules")
+                .takes_value(true)
+                .multiple(true)
+                .help("Selects modules to use"),
+        )
+        .arg(Arg::with_name("FILE").help("File to chisel"))
+        .arg(
```
axic: Don't wanna nitpick too much but would prefer to not include oneliner specific options in this PR, like `OUTPUT_PATH`, `MODULE_OPIONS` and `OUTPUT_MODE`.

chisel/src/main.rs
```diff
@@ -6,65 +6,102 @@ mod driver;
...
+        )
+        .arg(
+            Arg::with_name("MODULES")
+                .short("m")
+                .long("modules")
+                .takes_value(true)
+                .multiple(true)
+                .help("Selects modules to use"),
+        )
+        .arg(Arg::with_name("FILE").help("File to chisel"))
```
axic: This doesn't have a short or long option, does this mean it is a leaf argument?

axic: If so perhaps it should be last before the help message.

chisel/src/main.rs
```diff
@@ -6,65 +6,102 @@ mod driver;
...
                         .short("c")
                         .long("config")
-                        .help("Sets a custom configuration file")
-                        .value_name("CONF_FILE")
+                        .help("Sets the configuration file in config-driven mode.")
+                        .value_name("PATH")
                         .takes_value(true),
                 ),
         )
+        .after_help("chisel runs in two primary modes: unix-style and config-driven.\n\nunix-style is invoked without a subcommand. \
```
axic: Perhaps move this into the oneliner PR too.

chisel/src/main.rs
```diff
@@ -6,65 +6,102 @@ mod driver;
...
-            process::exit(chisel_run(subcmd_matches))
+        ("run", args) => {
+            if let Some(opts) = args {
+                flags.apply(opts);
+            }
+
+            chisel_run(flags)
         }
-        _ => err_exit(ERR_NO_SUBCOMMAND),
+        // Run chisel in unix mode if no subcommand provided
```
axic: Same here.

chisel/src/main.rs
```diff
@@ -6,65 +6,102 @@ mod driver;
...
-};
+use cmd_run::chisel_run;
+use options::ChiselFlags;
 
 fn fail(code: i32, message: &str) -> ! {
     eprintln!("{}: {}", crate_name!(), message);
     process::exit(code);
 }
 
-#[cfg(feature = "binaryen")]
```
axic: Hm was this a leftover?

chisel/src/main.rs
jakelang: It's an argument which is invoked without a flag. Will move to end.

chisel/src/main.rs
jakelang: Ok

chisel/src/main.rs
```diff
@@ -6,50 +6,79 @@ mod driver;
...
+        )
+        .arg(
+            Arg::with_name("DEBUG_MESSAGES")
+                .short("d")
+                .long("debug")
+                .help("Enables debug messages")
                 .global(true),
         )
+        .arg(
+            Arg::with_name("MODULES")
```
axic: Is this part of oneliner too ?

chisel/src/main.rs
jakelang: It's a description for both, I think this should be no big issue.

chisel/src/cmd_run.rs
```diff
@@ -1,315 +1,150 @@
...
+                Some("hex") => ruleset.write("hex"),
+                _ => panic!("CLI parser ensures value can only be one of the above"),
+            };
+            (ruleset_name, write_result)
+        })
+        .filter_map(|(name, result)| {
+            if result.is_err() {
+                Some((name, result.unwrap_err()))
+            } else {
+                //
```
axic: Oops, stray comment ?

PR #180, opened 2019-12-28
Some cleanups according to clippy

axic:


chisel/src/cmd_run.rs
```diff
@@ -110,7 +110,7 @@ pub fn chisel_run(flags: ChiselFlags) -> i32 {
     eprintln!("{}", &results);
     results
         .rulesets_mut()
-        .into_iter()
+        .iter_mut()
```
jakelang: I think I decided to explicitly consume by value here but it is a matter of personal preference.

PR #181, opened 2019-12-28
Introduce oneliner CLI mode

jakelang:


axic:
Can you update the README?

chisel/src/cmd_oneliner.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use crate::config::ChiselConfig;
+use crate::config::FromArgs;
+use crate::driver::{ChiselDriver, DriverState};
+use crate::fail;
+use crate::logger;
+use crate::options::ChiselFlags;
+
+pub fn chisel_oneliner(flags: ChiselFlags) -> i32 {
+    let log_level = match flags.value_of("util.debugging") {
```
axic: This seems to be duplicated in `cmd_run`. Should fix this in a separate PR.

jakelang: :+1:

chisel/src/cmd_oneliner.rs
```diff
@@ -0,0 +1,120 @@
...
+            chisel_debug!(1, "Modules passed:\n\t{}", module_list);
+
+            let options_list = if let Some(opts) = flags.value_of("oneliner.modules.options") {
+                chisel_debug!(1, "Module options passed:\n\t{}", opts);
+                opts
+            } else {
+                ""
+            };
+
+            // Input file
```
axic: Why not just name this `input` and remove the comments? (Same below)

chisel/src/cmd_oneliner.rs
```diff
@@ -0,0 +1,120 @@
...
+            // Output file
+            let output = flags.value_of("oneliner.output");
+            let output = match output {
+                Some(p) => p.to_string(),
+                None => "/dev/stdout".to_string(),
+            };
+
+            let chisel_config = match ChiselConfig::from_args(module_list, options_list) {
+                Ok(mut config) => {
+                    // Inject the input and output file paths here.
```
axic: Ugh. But oh well.

chisel/src/cmd_oneliner.rs
```diff
@@ -0,0 +1,120 @@
...
+            match io_result {
+                Ok(true) => eprintln!("Successfully wrote output to file."),
+                Ok(false) => eprintln!("No changes to write."),
+                Err(e) => {
+                    eprintln!("failed to write output to file: {}", e.description());
+                    exit_code = 1;
+                }
+            }
+
+            exit_code
```
axic: I think this could be removed just returning the code in the match? But at least moving the `exit_code` declaration in before this match. Where it is before the loop is too early.

chisel/src/cmd_oneliner.rs
jakelang: I know lol

PR #188, opened 2020-01-15
Use ModulePreset trait in TrimExports

axic:
When the trait was introduced possibly only the ExportWhitelist was updated.

### DIFF
--- a/libchisel/src/trimexports.rs
+++ b/libchisel/src/trimexports.rs
@@ -80,20 +80,6 @@ impl TrimExports {
         }
     }
 
-    /// Takes a given preset string and constructs a context with the
-    /// corresponding whitelist.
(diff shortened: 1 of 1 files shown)


libchisel/src/trimexports.rs
```diff
@@ -119,6 +105,22 @@ impl TrimExports {
     }
 }
 
+impl ModulePreset for TrimExports {
+    /// Takes a given preset string and constructs a context with the
+    /// corresponding whitelist.
+    fn with_preset(preset: &str) -> Result<Self, ModuleError> {
+        match preset {
+            "ewasm" => Ok(TrimExports {
+                whitelist: ExportWhitelist::with_preset("ewasm")?,
```
axic: Additionally this bubbles up errors now.

PR #193, opened 2020-01-15
Simplify create_memory_deployer

axic:
With using the with_data() helper of the builder


libchisel/src/deployer.rs
```diff
@@ -155,18 +155,14 @@ fn create_memory_deployer(payload: &[u8]) -> Module {
         // Add default memory section
         .memory()
             .with_min(memory_initial)
+            .with_data(0, payload.to_vec())
```
axic: Validated by changing this offset. The test failed with the offset being different.

PR #197, opened 2020-01-18
Remove new() from CheckFloat, Repack and Snip

axic:
Follow up of #161.
Depends on #198.


libchisel/src/dropsection.rs
```diff
@@ -3,7 +3,13 @@ use std::error::Error;
 
 use parity_wasm::elements::{Module, Section};
 
-use super::{ChiselModule, ModuleConfig, ModuleError, ModuleKind, ModuleTranslator};
+use super::{ChiselModule, ModuleError, ModuleKind, ModuleTranslator};
+
+impl From<std::num::ParseIntError> for ModuleError {
```
jakelang: Maybe we could just implement `From<E: Error> for ModuleError` so that we can use the ? operator globally. In the end it's all a `Custom` enum value.

PR #198, opened 2020-01-18
Merge ModuleConfig trait into ChiselModule

axic:


axic:
> but why not just scrap the old methods from ChiselModule?

You can once #172 is done.

